Issue #1, opened 2019-11-11
Code block missing in online doc


CharlyCast:
In "Getting Started" under "installing poke-env" the code block 

```python
pip install poke-env
```

Is not visible. Other blocks appears on the same page, and when building the doc locally I do get the pip install block. Any configuration issue?

![poke-env-doc](https://user-images.githubusercontent.com/23420125/68577832-2235fb00-0471-11ea-91ff-6695c2e90a8d.png)

---

EDIT:

After comparing the online doc with the one I built locally, I seems that many more code blocks are missing, for instance the instructions on how to setup a local showdown server:

### Online documentation:

![setup-live](https://user-images.githubusercontent.com/23420125/68578595-9a50f080-0472-11ea-805c-d51389ec4ab0.png)

### My locally build documentation

![setup-perso](https://user-images.githubusercontent.com/23420125/68578601-9c1ab400-0472-11ea-9896-3b312c23cdbe.png)

hsahovic:
Thank you for bringing that up.

I'll investigate an correct that asap.

hsahovic:
Fixed in #2.

PR #2, opened 2019-11-11
Specify block code types in index

hsahovic:
Fix for #1


docs/source/index.rst
```diff
@@ -22,7 +22,7 @@ Installing ``poke-env``
 
 Installation can be performed via pip with:
 
-.. code-block::
+.. code-block:: bash
```
CharlyCast: By curiosity, do you know why this is necessary? According to #1 building locally totally works.

PR #21, opened 2020-03-06
Explicit exception handling

hsahovic:


hsahovic:
The last dependabot BUMP broke CI 3.7. This needs to be fixed before further merges.

Issue #20, opened 2020-03-05
Gen 8 random battle results in no available moves


jzyrobert:
1. Run the performance showdown fork
2. Copy the random player tutorial but replace "gen7randombattle" with "gen8randombattle"
3. Run it, and it hangs until manually quit.

I've added print messages to the "choose_move", and an example returned object is:
`{'_battle_tag': 'battle-gen8randombattle-90', '_opponent_username': 'Player 1', '_player_role': 
'p2', '_player_username': 'Player 2', '_players': [{'username': 'Player 1', 'player': 'p1', 'avatar': '2', 'rating': ''}, {'username': 'Player 2', 'player': 'p2', 'avatar': '1', 'rating': ''}], 'logger': <Logger Player 2 (WARNING)>, '_available_moves': [], '_available_switches': [], '_can_mega_evolve': False, '_can_z_move': False, '_force_switch': False, '_in_team_preview': False, 
'_maybe_trapped': False, '_trapped': False, '_force_swap': False, '_wait': False, '_finished': False, '_rqid': 3, '_rules': ['Species Clause: Limit one of each Pokémon', 'HP Percentage Mod: HP is shown in percentages', 'Sleep Clause Mod: Limit one foe put to sleep'], '_turn': 1, '_won': 
None, '_weather': None, '_fields': set(), '_side_conditions': set(), '_opponent_side_conditions': set(), '_team': {'p2: Octillery': Octillery (pokemon object) , 'p2: Manectric': Manectric (pokemon object) }, '_opponent_team': {'p1: Glaceon': Glaceon (pokemon object) }}`

As you can see available_moves, available_switches are both empty.

I have also attempted to set log_level="DEBUG" for one player, and the output is [debug.txt](https://github.com/hsahovic/poke-env/files/4293101/debug.txt)

I will attempt to debug this, but do you have any idea where this might be going wrong (maybe in the message parsing?)

hsahovic:
Hi Robert,

Thanks for bringing that up. I will try to reproduce your issue this evening; I'll keep you posted. From the looks of it, it might be a gen8-related issue.

hsahovic:
The problem comes from gen8 mechanics / specifics. Right now, the project is not compatible with gen 8. However, gen8 compatibility is part of the roadmap.
It is planned as the next step after implementing an open ai gym wrapper, which is currently being built. If gen8 support is a priority for your project, feel free to reach out - it might be prioritized.

jzyrobert:
[...]

PR #26, opened 2020-03-25
Gen 8 support

szymonWojdat:
As discussed in #20

Copying over a checklist by @hsahovic [that he mentioned here](https://github.com/hsahovic/poke-env/issues/20#issuecomment-595553509) in order to keep track of what's left

Things to update:

- [x] src/data/pokedex.json
- [x] src/data/moves.json
- [x] src/environment/Battle.battle_parse_message
- [x] src/environment/Battle.battle_parse_request
- [x] src/player/Player.choose_random_move

Things to add:

- [x] src/environment/Battle.can_dynamax
- [x] src/environment/Pokemon._dynamax

Things to add if necessary:
- [ ] src/environment/Battle.can_gigantamax
- [ ] src/environment/Pokemon._gigantamax
- [x] src/environment/Pokemon.is_dynamaxed
- [ ] src/environment/Pokemon.is_gigantamaxed


szymonWojdat:
Merged `master` into `feature-gen-8-support` as it seemed like the correct thing to do, there weren't any conflicts, let me know if I did something stupid

hsahovic:
Yeah merging master makes sense. Let me know if you need a hand with anything :)

jzyrobert:
@szymonWojdat Thanks for picking this up, I was busy with exams, but let me know if you need any help.

About the removal of zmoves/mega, isn't that still needed if someone chooses to play an older generation?

hsahovic:
@jzyrobert This has been discussed with @szymonWojdat ; Megas and Z-Moves will be kept, and the library will remain compatible with gen 7 :)

szymonWojdat:
Alright got a little entangled in reverting, gonna reset the branch back to d8d82c4a045e6493c91b4f311fb09762c0c04fe2 and go from there

szymonWojdat:
Ok  i thought it would let me commit from d8d82c4 but it forced me to merge, so unfortunately my z-move/mega removing changes are still there, i'll sort this out tomorrow.

Btw currently working on parsing the dynamax-related requests/messages but kept most of the changes locally as it's still WIP and only pushed two tiny commits in order to reset the branch but unfortunately it doesn't work the way i thought and it merged the commits i wanted to get rid of, as mentioned above.

szymonWojdat:
`git push -f` does wonders. d8d82c4 is now `HEAD` of this branch.

alright so now that we got this sorted out, i'll be updating `moves.json` first as the simulation keeps failing if one of the mons has an unknown move, then I'll implement dynamax

szymonWojdat:
It is now possible to use dynamax but `Pokemon` state doesn't get updated yet - I'll work on that. So if you make two `RandomPlayer`s face off against each other, you should see them dynamax eventually.

szymonWojdat:
Hey, I just implemented `Pokemon._dynamax()` and `Pokemon._end_dynamax()`. The dynamax state is kept as an `Effect` in `Pokemon._effects`. It's possible to Dynamax a Pokemon through either `_dynamax()` or through `_start_effect()` (same with ending Dynamax) but I needed to duplicate some logic between those two methods because there is no public API for `Pokemon` class (the state gets modified from outside the class through protected methods, eg. in `Battle._handle_battle_message()` so had to assume that all protected methods can be used from outside the class. I guess an implementation of public `Pokemon` API will need to happen sooner or later.

Second thing is that I'm struggling to find where is `Battle._available_moves` getting modified? I'm asking because I was thinking about making only max moves available for already-dynamaxed Pokemon.

hsahovic:
> Hey, I just implemented Pokemon._dynamax() and Pokemon._end_dynamax(). The dynamax state is kept as an Effect in Pokemon._effects. It's possible to Dynamax a Pokemon through either _dynamax() or through _start_effect() (same with ending Dynamax) but I needed to duplicate some logic between those two methods because there is no public API for Pokemon class (the state gets modified from outside the class through protected methods, eg. in Battle._handle_battle_message() so had to assume that all protected methods can be used from outside the class.

How is the effect managed in `Battle._handle_battle_message`? Is the information that a Pokemon is dynamaxing transmitted as an effect (eg. `-activate` message)?

If that's the case, we maybe do not need a `_dynamax` method, and directly handle side effects in `_start / _end_effect`.

Otherwise, we could use a dedicated method and a boolean, similarly to what's done for mega-evolutions.

> Second thing is that I'm struggling to find where is Battle._available_moves getting modified? I'm asking because I was thinking about making only max moves available for already-dynamaxed Pokemon.

That would be in `Battle._parse_request` :)

A couple of questions:

- Is the HP always multiplied by 2 in PS?
- Would it be cleaner to explicitly multiply both `current_hp` and `max_hp` on dynamax?
- Should we think of shedinja?
- Make sure that dynamax effects are reinitialized on switch out :)
- We should track dynamaxes in `Battle` objects, so we can tell if we / the opponent can dynamax

szymonWojdat:
> How is the effect managed in Battle._handle_battle_message? Is the information that a Pokemon is dynamaxing transmitted as an effect (eg. -activate message)?

Dynamax is managed in `Battle._parse_message` via `-start` and `-end` messages.

> If that's the case, we maybe do not need a _dynamax method, and directly handle side effects in _start / _end_effect.

I'm just wondering how to we handle dynamax from the `Player` API perspective? Like the agent needs to somehow be able to dynamax whenever it wants. Do we just make it possible through selecting a max move?

> That would be in Battle._parse_request :)

ohh I just found it thanks :) it gets modified via a **getter**, as it returns a reference to the protected attribute (list), this seems a bit hacky, was this intended or should I make it modify the protected attribute directly?

> Is the HP always multiplied by 2 in PS?

I think so, I can double check in their codebase

> Would it be cleaner to explicitly multiply both current_hp and max_hp on dynamax?

agreed

> Should we think of shedinja?

yes, good point!

> Make sure that dynamax effects are reinitialized on switch out :)

oh right good point

> We should track dynamaxes in Battle objects, so we can tell if we / the opponent can dynamax

alright gonna look into that

hsahovic:
Thanks for your detailed answer :)

> Dynamax is managed in Battle._parse_message via -start and -end messages.
 
Ok; if that is the case I do not think we should have separate methods to manage dynamax, as it can be seen as a standard effect; however we do have to take care of HPs and make sure that the effect disappears when effects are cleared / the pokemon swttches out.

> I'm just wondering how to we handle dynamax from the Player API perspective? Like the agent needs to somehow be able to dynamax whenever it wants. Do we just make it possible through selecting a max move?

I think that this should be a question to be handled by returning orders while selecting moves; `Player.create_order` allows options for mega evolution and z-moves (its signature is currently `order: Union[Move, Pokemon], mega: bool = False, z_move: bool = False`). Dynamaxing should be a similar option :)

> ohh I just found it thanks :) it gets modified via a getter, as it returns a reference to the protected attribute (list), this seems a bit hacky, was this intended or should I make it modify the protected attribute directly?

That was not intended; feel free to correct it :)

> I think so, I can double check in their codebase

Perfect

> Should we think of shedinja?
> yes, good point!

There might be other edge cases like this; it's not a priority but if you think of some please include them :) Asking on the discord might be an idea

Thanks @szymonWojdat, you're doing a super job :)

On a side note, the tests are not passing right now - which makes sense considering the number of changes introduced. They will need to pass before merging; if you need a hand feel free to reach out :)

szymonWojdat:
I noticed we're duplicating work by tracking available moves (potentially `can_dynamax` and so on once implemented) separately in `Battle` and in `Pokemon`. How do you feel about something like that going forward?

```python3
@property
def available_moves(self) -> List[Move]:
    """
    :return: The list of moves the player can use during the current move request.
    :rtype: List[Move]
    """
    return self.active_pokemon.available_moves
```

hsahovic:
I don't think that we are tracking `available_moves` in `Pokemon`. We have `moves`, but all moves are not available: the idea behind `available_moves` being in `Battle` is that this is updated with every request to take into account the state of the game, including the active pokemon and other special cases, such a recharging a move, regardless of the current state of the underlying `Pokemon` object, while the `Pokemon` object simply keeps track of the moveset, both for our and the opposing pokemons.

If you think that keeping track of available moves directly in the `Pokemon` instance would be better, I suppose we could do that, but I'd like to be convinced before jumping into it :)

szymonWojdat:
Alright, now that I think about it, this information (`can_dynamax`) actually belongs in the `Battle` state because it tells whether it is possible to dynamax or not, while `Pokemon` instance only knows if it's *currently* dynamaxed or not. When it comes to `available_moves` I think I'll just keep it simple for now. I was considering adding it to `Battle` state to have it dynamically change between max moves, regular moves and both depending on the dynamax state of the `Battle` (which would also make `Player.choose_random_move()` a bit simpler. But for now I think it's going to be the best to do it in a similar way to max moves/megas and maybe optimize later on. Of course I won't be making such changes without discussing them with you first.

szymonWojdat:
I just managed to rerun tests locally, adjusted a few methods, nothing too big, just pushed out the changes. A couple of issues:
* `Move.can_z_move` doesn't work as expected as `moves.json` no longer contains `zMovePower` for all damage moves, my guess is that PS has switched to some sort of dynamic calculation of the z-move power. The only place where it's used in code is at `Pokemon.available_z_moves()`. I assume the easiest workaround would be to modify `Move.can_z_move` so that it checks if given moves appears on a small list of moves that cannot z-move (`test_can_z_move()` implies as if Metronome could not z-move but it actually can, it just doesn't have any extra effect, moves that can't z-moves would be eg. Struggle)
* for the same reason as mentioned above, `test_z_move_power` is failing as well. I'll try taking a look at their codebase to see what's the formula that they're using

hsahovic:
@szymonWojdat thanks for bringing that up, I'll look into it :)

szymonWojdat:
Just added:
* `Battle.opponent_can_dynamax` - as discussed
* `Battle.dynamax_turns_left` - as a way of determining how many turns of dynamax are left for player, I figured this would be crucial information for any agent as it wouldn't have a way of figuring out how many turns of dynamax are left otherwise based on the current `Battle` state
* `Battle.opponent_dynamax_turns_left` - same as above but for the opponent

I have tested both locally manually. Aside from writing tests for the new functionalities and fixing the failing ones, I believe we're done?

hsahovic:
We'll have to update docs too and be sure that z-moves still work as expected. Other than that, yes :)

szymonWojdat:
Just added a simple test for dynamax. Do you mind explaining what do you mean by "be sure that z-moves work as expected"? Were you refering to the fact that those were gone for a while and then moved back in? If so, then those commits aren't even in the tree anymore, so I'd expect them to work normally. I can also test those manually against my local PS instance or write a simple unit test if necessary.

hsahovic:
I was referring to this:
>  -  Move.can_z_move doesn't work as expected as moves.json no longer contains zMovePower for all damage moves, my guess is that PS has switched to some sort of dynamic calculation of the z-move power. The only place where it's used in code is at Pokemon.available_z_moves(). I assume the easiest workaround would be to modify Move.can_z_move so that it checks if given moves appears on a small list of moves that cannot z-move (test_can_z_move() implies as if Metronome could not z-move but it actually can, it just doesn't have any extra effect, moves that can't z-moves would be eg. Struggle)
>  -  for the same reason as mentioned above, test_z_move_power is failing as well. I'll try taking a look at their codebase to see what's the formula that they're using

szymonWojdat:
Thanks! I fixed both of those. Also added two `pyre-ignore`s.

Now integration tests are failing. `test_max_concurrent_battle_works` is timing out. Do you mind taking a look? I guess we'll also need to add an integration test for `gen8randombattle`

hsahovic:
@szymonWojdat Yes, sure! I just need to finish a couple of uni-related things; I'll get back into `poke-env` by the end of the week

PR #42, opened 2020-05-05
Feature gen 8

hsahovic:


hsahovic:
Note: investigate the effects of healing while dynamaxing

szymonWojdat:
> Note: investigate the effects of healing while dynamaxing

I believe PS! just sends a `-heal` message with an exact HP number that given Pokemon's HP needs to be set to (or HP percentage, in case of the opponent's Pokemon).

hsahovic:
Yes, the problem I have in mind is related to that happening with an opponent dynamaxed mon. Iirc, receiving a heal messages updates the mon's hp figures, which could conflict with our manual tickling of these with dynamax.
For instance, you could end up with an opposing mon having a counter out of 50.

hsahovic:
On the hp things: after investigating a bit, showdown sends a `-heal` message containing information on updated hp after dynamaxing. We do not have to deal with it ourselves, which simplifies things quite a lot.

I'm going to revert back this part of the code.

PR #51, opened 2020-05-20
Doubles support

### LINKED ISSUE #49: Doubles support
I started messing around with doubles and managed to successfully play a couple of turns but before I proceed with this, I think it's important to decide in which direction do we want to go with this design-wise as there's a couple options:

1. Create `DoubleBattle` class that extends `Battle`, same with `DoublesPlayers` that extends `Player`
2. Make backwards-incompatible changes to `Player` and `Battle` classes so that they support both singles at doubles at the same time. This seems like a more elegant way but will involve more work and testing (which I'm fine with). Eg. `Battle.active_pokemon` will need to return an iterable of `Pokemon` instances rather than one instance and so on.

Please let me know what do you think @hsahovic

szymonWojdat:
Closes #49 

Drafted this real quick for now, will keep working on this in the following days.

For now the plan was to create `AbstractBattle`, use `DoubleBattle` for doubles and `Battle` for singles, move all common logic from `Battle` to `AbstractBattle` and implement the missing functionalities of `DoubleBattle`.


szymonWojdat:
I think most of the work on `DoubleBattle` is complete, stuff that's left:
* isolating a `DoublesPlayer` class as I think it's the cleanest way of integrating `DoubleBattle` with `Player`. Similarly to `DoubleBattle`, it will inherit from an abstract class, `Player` class will do so too
* implementing `DoublesPlayer.choose_random_move()`
* writing a ton of unit tests for doubles
* creating some simple baselines for doubles (?)

szymonWojdat:
Also I don't quite understand why is flake8 (line length over 88) failing for `double_battle.py` while it clearly should be failing for other files as well (eg. `battle.py`).

hsahovic:
Good job! Are you using `black`? I suspect it might solve most of these issues (apart from docstrings, which might need to be manually reformatted).

szymonWojdat:
@hsahovic thanks for the advice, gonna use it from now on, for some reason I've never noticed that max line length was so small.

Just fixed all tests and stuff, going to work on `DoublesPlayer` from tomorrow. Mind taking a look at the last step that keeps failing? Looks like there's some issue on the CI machine, does it not reset its state after each run?

hsahovic:
Oh, I think that I know what the problem is. I updated the CI process on another branch last week, and I think that my first updates might have used the same cache name - which was an error on my end. I think that merging the current state of master into this branch should solve the issue.

On the subject of tests, you should also install `pre-commit`: it will help you catch this kind of errors before committing your work.

szymonWojdat:
@hsahovic the build is failing and the stack trace contains a reference to a non-existent line of code
```
  File "/home/circleci/repo/venv/lib/python3.8/site-packages/poke_env/environment/battle.py", line 177, in active_pokemon
    raise ValueError("No active pokemon found in the current team")
```
Context - it kept raising the error in case a Pokemon fainted to Stealth Rocks, when it tried switching a Pokemon in and replacing an active Pokemon, but the active Pokemon didn't exist as no Pokemon was marked active, so it raised the ValueError. I successfully fixed it locally by making `Battle.active_pokemon` return `Optional[Pokemon]` instead of `Pokemon` and adjusted the rest of the codebase accordingly.

Can you please advise? Thanks

hsahovic:
@szymonWojdat Sorry about that - that's indeed a caching issue we uncovered with @dmizr last week and fixed on master.

You can fix it by updating `.circleci/config.yaml` from [this version](https://github.com/hsahovic/poke-env/blob/master/.circleci/config.yml) and updating the version identifier line 40 and 143 to (I'd recommend `-v3` since we used `-v2` already) 

The issue stems from a couple of modifications I made last month to accelerate the ci, which accidentally cached a previous version of `poke-env` which is being restored and used instead of your current code.

szymonWojdat:
> @szymonWojdat Sorry about that - that's indeed a caching issue we uncovered with @dmizr last week and fixed on master.
> 
> You can fix it by updating `.circleci/config.yaml` from [this version](https://github.com/hsahovic/poke-env/blob/master/.circleci/config.yml) and updating the version identifier line 40 and 143 to (I'd recommend `-v3` since we used `-v2` already)
> 
> The issue stems from a couple of modifications I made last month to accelerate the ci, which accidentally cached a previous version of `poke-env` which is being restored and used instead of your current code.

thanks! this seems to have worked

szymonWojdat:
I've just found one more issue with `DoubleBattle.active_pokemon` - `DoubleBattle._active_pokemon` does not get properly updated with each request, so far I've updated it only on switch. I need to investigate a bit more, then I'll go ahead and write the tests that we discussed.

szymonWojdat:
@hsahovic "it works on my machine" - i believe CIrcleCI is pulling an older version of `double_battle.py` from cache since i updated it recently with 4c408e3. Do you mind taking a look????

hsahovic:
Hey @szymonWojdat ! I can take a look at it tonight. Alternatively, you may invalidate the cache by changing it's version in `.circleci...`

szymonWojdat:
> Hey @szymonWojdat ! I can take a look at it tonight. Alternatively, you may invalidate the cache by changing it's version in `.circleci...`

I think it's my bad - it's clearly pulling my latest code and the repo code is not getting cached by CI, at least judging by CI config, so this cannot be a cache issue. I'll keep investigating

dmizr:
Hey @hsahovic @szymonWojdat, I noticed some effects that occur in double battles are absent from the `src/environment/effect.py` file. 
Here's a version of the `Effect` class with several extra effects that can happen in battles: https://gist.github.com/dmizr/1f1d6911ea4cc20203dbd0d73f417f49

szymonWojdat:
@hsahovic i have no idea why flake is failing, should i reformat docstrings in `excetptions.py` so that they're even shorter?

hsahovic:
I'll check it out tonight.

PR #52, opened 2020-05-29
Ci and process improvement

hsahovic:


hsahovic:
Tests for the last commit failed because of a bug related to ditto. This require additional investigation.

PR #138, opened 2021-04-07
Generation specific data (Gen 4 - Gen 8)

AFumis:
Creates generation specific child classes for Battle, Pokemon and Move, which use data that correctly reflects the games from the generation, including but not limited to:
- Move base power and accuracy
- Typing
- Abilities
- Base stats


hsahovic:
Thanks for the PR! Can you make sure it passes circle ci's tests? I'll review it when it's done :)

PR #180, opened 2021-07-18
Check for items & abilities when parsing damage & heal messages

pmariglia:
Hey Haris :wave: 

Been using poke-env for some personal testing and noticed some information not being parsed from the battle-log.

This PR parses and updates the battle object for

**items** from:
- the item-holder damaging the opponent (example: Rocky Helmet)

**abilities** from:
- the ability healing the pokemon to which it belongs (example: Water Absorb, Volt Absorb)
- the ability damaging the opponent's pokemon (example: Iron Barbs, Rough Skin)

I've tried to write this in a way that I think makes sense with your project - but feel free to criticize anything and everything I wrote :upside_down_face: 

Items healing the pokemon to which they belong is intentionally omitted since they may be consumed during the process (berries), but the message for consumption happens _before_ the -heal message so this would result in an incorrect state. I have a solution for this that I've used in my project but I'll leave that for another discussion/PR.

### DIFF
--- a/src/poke_env/environment/abstract_battle.py
+++ b/src/poke_env/environment/abstract_battle.py
@@ -210,6 +210,54 @@ class AbstractBattle(ABC):
     def _clear_all_boosts(self):  # pragma: no cover
         pass
 
+    def _check_damage_message_for_item(self, split_message: List[str]):
+        # Catches when a side takes damage from the opponent's item
+        # The item belongs to the side not taking damage
+        # Example:
+        #   |-damage|p2a: Archeops|88/100|[from] item: Rocky Helmet|[of] p1a: Ferrothorn
+        if (
+            len(split_message) == 6
+            and split_message[4].startswith("[from] item:")
+            and split_message[5].startswith("[of]")
+        ):
+            item = split_message[4].split("item:")[-1]
+            pkmn = split_message[5].split("[of]")[-1].strip()
+            self.get_pokemon(pkmn).item = to_id_str(item)
+
+        # Catches when a side takes damage from it's own item
+        # The item belongs to the same side taking damage
+        # Example:
+        #   |-damage|p2a: Pikachu|90/100|[from] item: Life Orb
+        elif len(split_message) == 5 and split_message[4].startswith("[from] item:"):
+            item = split_message[4].split("item:")[-1]
+            pkmn = split_message[2]
+            self.get_pokemon(pkmn).item = to_id_str(item)
+
+    def _check_damage_message_for_ability(self, split_message: List[str]):
+        # Catches when a side takes damage from the opponent's ability
+        # the item is from the side not taking damage
+        # Example:
+        #   |-damage|p2a: Archeops|88/100|[from] ability: Iron Barbs|[of] p1a: Ferrothorn
+        if (
+            len(split_message) == 6
+            and split_message[4].startswith("[from] ability:")
+            and split_message[5].startswith("[of]")
+        ):
+            ability = split_message[4].split("ability:")[-1]
+            pkmn = split_message[5].split("[of]")[-1].strip()
+            self.get_pokemon(pkmn).ability = to_id_str(ability)
+
+    def _check_heal_message_for_ability(self, split_message: List[str]):
+        # Catches when a side heals from it's own ability
+        # the "of" component sent by the PS server is a bit misleading
+        #   it implies the ability is from the opposite side
+        # Example:
+        #   |-heal|p2a: Quagsire|100/100|[from] ability: Water Absorb|[of] p1a: Genesect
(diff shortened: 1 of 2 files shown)


unit_tests/environment/test_battle.py
```diff
@@ -405,6 +405,56 @@ def test_battle_request_and_interactions(example_request):
     assert not battle.maybe_trapped
     assert battle.opponent_can_dynamax
 
+    # Items
+    battle._parse_message(
+        ['', '-damage', 'p2a: Necrozma', '167/319', '[from] item: Rocky Helmet', '[of] p1a: Groudon']
+    )
+    assert battle.opponent_active_pokemon.item == "rockyhelmet"
+    battle.opponent_active_pokemon._item = None
```
pmariglia, line 413: This is one long test, so I reset the item/ability changes like this to make sure a later test doesn't assert something that happened in a previous test. Need to be careful when asserting multiple things in one test :slightly_smiling_face:

src/poke_env/environment/abstract_battle.py
```diff
@@ -210,6 +210,54 @@ def get_pokemon(
...
+        # Example:
+        #   |-damage|p2a: Archeops|88/100|[from] item: Rocky Helmet|[of] p1a: Ferrothorn
+        if (
+            len(split_message) == 6
+            and split_message[4].startswith("[from] item:")
+            and split_message[5].startswith("[of]")
+        ):
+            item = split_message[4].split("item:")[-1]
+            pkmn = split_message[5].split("[of]")[-1].strip()
+            self.get_pokemon(pkmn)._item = to_id_str(item)
```
hsahovic, line 225: `Pokemon` objects have a setter method associated with the `item` property - could you do

```python
self.get_pokemon(pkmn).item = to_id_str(item)
```
instead? For now it's strictly equivalent, but if might be used for other callbacks at some point.

src/poke_env/environment/abstract_battle.py
```diff
@@ -210,6 +210,54 @@ def get_pokemon(
...
+
+        # Catches when a side takes damage from it's own item
+        # The item belongs to the same side taking damage
+        # Example:
+        #   |-damage|p2a: Pikachu|90/100|[from] item: Life Orb
+        elif len(split_message) == 5 and split_message[4].startswith("[from] item:"):
+            item = split_message[4].split("item:")[-1]
+            pkmn = split_message[2]
+            self.get_pokemon(pkmn)._item = to_id_str(item)
```
hsahovic, line 234: Same thing as above :)

src/poke_env/environment/abstract_battle.py
```diff
@@ -210,6 +210,54 @@ def get_pokemon(
...
+        # Example:
+        #   |-damage|p2a: Archeops|88/100|[from] ability: Iron Barbs|[of] p1a: Ferrothorn
+        if (
+            len(split_message) == 6
+            and split_message[4].startswith("[from] ability:")
+            and split_message[5].startswith("[of]")
+        ):
+            ability = split_message[4].split("ability:")[-1]
+            pkmn = split_message[5].split("[of]")[-1].strip()
+            self.get_pokemon(pkmn)._ability = to_id_str(ability)
```
hsahovic, line 248: `ability` also has a setter - I'd change it to `self.get_pokemon(pkmn).ability = to_id_str(ability)`

src/poke_env/environment/abstract_battle.py
```diff
@@ -210,6 +210,54 @@ def get_pokemon(
...
+    def _check_heal_message_for_ability(self, split_message: List[str]):
+        # Catches when a side heals from it's own ability
+        # the "of" component sent by the PS server is a bit misleading
+        #   it implies the ability is from the opposite side
+        # Example:
+        #   |-heal|p2a: Quagsire|100/100|[from] ability: Water Absorb|[of] p1a: Genesect
+        if len(split_message) == 6 and split_message[4].startswith("[from] ability:"):
+            ability = split_message[4].split("ability:")[-1]
+            pkmn = split_message[2]
+            self.get_pokemon(pkmn)._ability = to_id_str(ability)
```
hsahovic, line 259: Same as above

hsahovic (commented):
Thanks a lot for this PR - it's a clear improvement. The only thing that I'd like to change is the setter property things, but it's really minor.
The integration tests are broken because of a typescript change in PS - I'll try to get a fix today, and release it with #176

pmariglia:
> Thanks a lot for this PR - it's a clear improvement. The only thing that I'd like to change is the setter property things, but it's really minor.
> The integration tests are broken because of a typescript change in PS - I'll try to get a fix today, and release it with #176

Fixed.

I had an earlier version of the project checked-out and was just copying what was done elsewhere :sweat:. Good to see it's now changed everywhere. :+1:

PR #186, opened 2021-07-31
Parse messages from items healing a Pokemon

pmariglia:
An addition to e93835abb1419bc28fb9e2f342ab563f5428ac04. This allows the
environment to parse items that heal the holder: Most notably Leftovers
and berries.

This also required an UNKNOWN_ITEM constant that each Pokemon's item
will now be initialized as, allowing the environment to distinguish between no item
(None) and an unknown item. This is necessary because the PS server will
give a log that looks like this when a berry is consumed:

```
|-damage|p2a: Blissey|22/714
|-enditem|p2a: Blissey|Sitrus Berry|[eat]
|-heal|p2a: Blissey|200/714|[from] item: Sitrus Berry
```

In the above log, the berry has already been consumed, but the heal
statement comes afterwards. The item will be None in this situation
because the `-enditem` log will have set it to None.

Making UNKNOWN_ITEM a string constant seemed logical, but whatever you think is appropriate is fine.

### DIFF
--- a/src/poke_env/environment/abstract_battle.py
+++ b/src/poke_env/environment/abstract_battle.py
@@ -248,6 +248,20 @@ class AbstractBattle(ABC):
             pkmn = split_message[5].split("[of]")[-1].strip()
             self.get_pokemon(pkmn).ability = to_id_str(ability)
 
+    def _check_heal_message_for_item(self, split_message: List[str]):
+        # Catches when a side heals from it's own item
+        # the check for item is not None is necessary because the PS simulator will
+        #  show the heal message AFTER a berry has already been consumed
+        # Examples:
+        #  |-heal|p2a: Quagsire|100/100|[from] item: Leftovers
+        #  |-heal|p2a: Quagsire|100/100|[from] item: Sitrus Berry
+        if len(split_message) == 5 and split_message[4].startswith("[from] item:"):
+            pkmn = split_message[2]
+            item = split_message[4].split("item:")[-1]
+            pkmn_object = self.get_pokemon(pkmn)
+            if pkmn_object.item is not None:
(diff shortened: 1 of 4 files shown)


hsahovic:
Thanks a lot! I'll review these two PRs tonight.
Best,
Haris

src/poke_env/environment/abstract_battle.py
```diff
@@ -248,6 +248,20 @@ def _check_damage_message_for_ability(self, split_message: List[str]):
             pkmn = split_message[5].split("[of]")[-1].strip()
             self.get_pokemon(pkmn).ability = to_id_str(ability)
 
+    def _check_heal_message_for_item(self, split_message: List[str]):
+        # Catches when a side heals from it's own item
+        # the check for item is not None is necessary because the PS simulator will
+        #  show the heal message AFTER a berry has already been
```
hsahovic, line 254: I think you are missing a word here :)

hsahovic (commented):
This looks good - I'd just like to move the definition of the constant before merging

src/poke_env/environment/pokemon.py
```diff
@@ -16,6 +16,9 @@
 from poke_env.utils import to_id_str
 
 
+UNKNOWN_ITEM = "unknown_item"
```
hsahovic, line 19: The string definition works fine. Can you move it to `poke_env.data`?

src/poke_env/environment/abstract_battle.py
pmariglia: :man_facepalming:

hsahovic:
Thanks a lot! This is good to merge.
I'll also solve #187 and publish these in the next release :)

Issue #208, opened 2021-09-28
Rework default player configuration to support parallel runs


Benjamin-Etheredge:
The current default player configuration generator will cause errors when two runs are kicked off. This is likely due to collisions with the default configuration counter.

hsahovic:
Hey @Benjamin-Etheredge,

That sounds like a very good catch. Do you have a code snippet I could use to reproduce it?

Benjamin-Etheredge:
I have a fix for it. I'm trying to get it pushed to a new branch now.

Benjamin-Etheredge:
I don't have a code snippet to reproduce it. I noticed it when trying to run the same code twice in two shells.

Benjamin-Etheredge:
There may also be an issue with passing specific player configurations through extended PlayerEnvs. I'm still investigating that one as it could be an issue with my code. It was easier to just rewrite the default configuration generator rather than investigate why my custom player names weren't working. The rework now works for parallel runs though.

hsahovic:
I don't think that `uuid`s is the way to go - using the counter-based solution creates a clearer sequence of player usernames, which can be useful when debugging or running certain types of algorithms (eg. genetic algorithms with a clear number / generation correlation).
If you can provide a snippet to reproduce this issue, I'll gladly issue a fix (probably using a `Lock` object).

Benjamin-Etheredge:
I'm not sold on uuid either. It is ugly and hard to track. I had to start logging player names to follow what was going on.
I like the more sequential nature, but it leads to collisions when two default configurations are used. 

I reproduced it by making a python file with the example code (heads up, there is a typo in the indentions of that example code too)
```
# -*- coding: utf-8 -*-
import asyncio
import time

from poke_env.player.player import Player
from poke_env.player.random_player import RandomPlayer


class MaxDamagePlayer(Player):
    def choose_move(self, battle):
        # If the player can attack, it will
        if battle.available_moves:
            # Finds the best move among available ones
            best_move = max(battle.available_moves, key=lambda move: move.base_power)
            return self.create_order(best_move)

        # If no attack is available, a random switch will be made
        else:
            return self.choose_random_move(battle)


async def main():
    start = time.time()

    # We create two players.
    random_player = RandomPlayer(
        battle_format="gen8randombattle",
    )
    max_damage_player = MaxDamagePlayer(
        battle_format="gen8randombattle",
    )

    # Now, let's evaluate our player
[...]

PR #210, opened 2021-09-29
Swapped default configuration to use uuid

Benjamin-Etheredge:
The goal is to fix collisions in player names across parallel runs.
There could still be collisions across runs, but it's not likely.


Benjamin-Etheredge:
Addresses #208

Benjamin-Etheredge:
Stripping hyphens from uuid would be a good idea before adding to the name string.

Issue #207, opened 2021-09-28
Feature: Different exposing of gym environment


Benjamin-Etheredge:
Have the code base register a gym environment. I recently saw a codebase that seemed to register its environment with gym.
This appears simple to do in the code base. That way anyone who installs/imports `poke-env` will be able to create a battler with `gym.make(...)`.

This would require a few things.

1. A showdown server already running somewhere
2. A rewrite of the PlayerEnv or a new PlayerEnv. 
3. Registering with gym through somewhere in the codebase. 

I have some ideas for a PlayerEnv rewrite. The basic idea is to have PlayerEnv keep track of opponents and opponent IDs. When given an opponent, that opponent will be set to just accept all challenges. This allows the player to just challenge whatever opponent ID whenever it wants. 
Default opponents can also be set such as base PlayerEnvs for random or the baseline models that exist in the codebase.

This is to facilitate easier access to a gym wrapper as well as add support for self-play. By simply pushing more opponents onto the list, different opponents can be cycled every game. A sample method can also be added to allow flexibility in how opponents are challenged and in what order.

I know this description likely isn't very clear. I'm working on implementing it now. I'll try to get some preliminary code in a pull request soon for feedback.

hsahovic:
@Benjamin-Etheredge yeah, both this and deep `PlayerEnv` rewrites are planned, but the current plan depends on other deep backend changes (making things more synchronous in general).
I was thinking of maybe having something like a `PlayerEnv.set_opponent` method or even `PlayerEnv.opponent` attribute, that would default to a `RandomPlayer` - do you think that would be a good api?

Benjamin-Etheredge:
I think going more synchronous is probably the way to go. Maybe just throw things on other processes? I'm not super familair with python multi-threading/processing.

I sorta did something similar. I had an opponent with a `set_policy` method. Then had the learner call that to update it. Something like having a `set_opponent` was something I considered. It would probably work just fine. 

For self-play stuff, I'm debating whether the env should keep track of which opponents are played and which are in the bank of opponents. By letting the environment keep track of it, the learner still has to call `add_opponent`. That's not much different than just letting the learner track opponents and call `set_opponent` though.

So I'm leaning more towards just having a `set_opponent` or `set_opponent_policy` type thing. I'm not totally sure what the best way is yet though.

[...]

PR #213, opened 2021-09-29
Exploring DummyPlayer usage

Benjamin-Etheredge:


Benjamin-Etheredge:
Still trying to figure out the async stuff for sending and accepting challenges automatically.

Benjamin-Etheredge:
Exploring #207

Benjamin-Etheredge:
While I was able to get gym registering different environments so I could just call something like `gym.make("RandomSingles-v8")`, I could not get it to work in parallel. Something hangs during `reset()` when two environments are created. I'm working on an example case for replicating it. 

Getting the ability to run multiple environments in parallel is really needed for faster training through `VecEnvs` like in OpenAI Baselines.

I'm still debugging it, but I'm a super novice with the async paradigm.

hsahovic:
@Benjamin-Etheredge to get multiple envs in parallel for training you'll get multiple instances of `PlayerEnv` as it is currently set-up to have one active battle at a time. That doesn't prevent having one shared model being trained though.

Benjamin-Etheredge:
@hsahovic something is going on with multiple instances of `PlayerEnv`. For some reason, everything breaks if I instantiate 2 `PlayerEnv` objects. It creates all 4 players and starts games for 2 envs but hangs on `reset` calls.

I'm not sure if it's an issue with the underlying implementation with async stuff or just my extra code. I'm attempting to mimic your implementation of putting the algorithm logic on a separate thread and running the event loop on the main thread. No dice yet.

PR #216, opened 2021-10-05
Init more exploration

Benjamin-Etheredge:


Benjamin-Etheredge:
@hsahovic I'm failing horribly at getting multiple environments to work. Asyncio is killing me. It is definitely not my forte. I'm going down the route of creating environments on different processes that each start their own asyncio loop (`SubprocVecEnv`) No dice. I can get it working for 1 environment by throwing the event loop on another thread, but it never works for more than 1 environment.

The goal is to get it started with playing an opponent in the `__init__`. If that works, it's trivial to wrap it to allow `gym.make("RandomSingles-v8")` to work. Again, it only works for creating a single environment this way. Any `gym.make()`/`CustomeEnvPlayer()` crashes things.

This is the kind of stuff I'm trying to adapt: https://stable-baselines3.readthedocs.io/en/master/guide/quickstart.html
Specifically: https://stable-baselines3.readthedocs.io/en/master/guide/vec_envs.html

I've got some ideas for how to rework `EnvPlayer` to be more "gym-like" and extendable, but the ability to run multiple agents at once is way more important (I think).

hsahovic:
Thanks for putting the work! I'll take a deeper look at what is doing on here later this week. As you can see, the whole `async` situation is a bit thorny 😅 I agree with you though - making this easy to scale would be great!

Benjamin-Etheredge:
No problem! I just REALLY want to do pokemon battles 😅 

It's super reassuring to hear you call it thorny too.

Benjamin-Etheredge:
Also, I *may* have just gotten it to work... It's looking promissing...

Benjamin-Etheredge:
It's working but seems bottlenecked by the inability to run in multiprocessing mode due to overlap in player names (i.e., each process tries to create `Player1`. I'm reimplementing the `uuid` stuff to see if it's faster.

hsahovic:
Regarding player names, you can specify which name to give a `Player` when instantiated - would that be a solution to this problem?

PR #217, opened 2021-10-06
Extensive interface reworks

Benjamin-Etheredge:


Benjamin-Etheredge:
@hsahovic These are some draft changes I was thinking of for reworks.

See how inheritance is replaced with composition for better modularity. See `src/poke_env/__init__.py` for how the components would fit together in a `gym.register(...)` call. Note: none of it works yet. I've still got some debugging to do on why those changes broke the async stuff (again).

Users could then extend these environments with their own `gym.register` calls to make whatever they want.

Benjamin-Etheredge:
Side note, it might be worth adding a showdown docker-compose folder so users can quickly and easily bring up their own showdown servers. That's what I'm using and it's super convenient due to showdown bugging out if not restarted properly or not having a `logs` folder or not having enough workers now that it's being blasted with tons of agents.

Benjamin-Etheredge:
It actually needs another rework.

The `EnvPlayer` environment should just return a reward of 0 and a `Battle` object. Then a wrapper should be used to generate a `BaseEnv-v8` type environment using `reward_computer` and `battle_embeder` or some combination. That way users would get direct access to the data and would encode it, stack it, or update reward signals entirely through gym wrappers which seems to be the convention for gym environments.

hsahovic:
Do you have docs related to the use of wrappers?
I never had any issue with showdown being overloaded / needing restarts while doing ~20 battles per second - did you increase its resource allocation?

PR #240, opened 2021-11-29
fix UnicodeEncodeError: 'charmap' codec can't encode character '\u260…

DrDamnation:
…6' in position 5319: character maps to <undefined> error


DrDamnation:
@hsahovic

PR #248, opened 2022-01-04
Fix deepcopy recursion

MatteoH2O1999:
Allows to deepcopy the battle object avoiding '\_\_setstate\_\_' hook from copy library to create a loop as self._parent is not initialised. This allows to study history-aware solutions.


hsahovic:
Thanks @MatteoH2O1999 !

PR #259, opened 2022-02-09
New player class for OpenAI Gym API with player async events in background thread

MatteoH2O1999:
To ensure full compliancy with OpenAI Gym API, the entire async process needs to happen in a background thread. These are the first steps in that direction. The player works as expected. Need to test more extensively and improve cleanup but it looks promising.

### DIFF
--- a/integration_tests/test_double_battles.py
+++ b/integration_tests/test_double_battles.py
@@ -28,18 +28,19 @@ async def simple_cross_evaluation(n_battles, format_, i=0):
     await cross_evaluate(players, n_challenges=n_battles)
 
     for player in players:
+        player.reset_battles()
         await player.stop_listening()
 
 
 @pytest.mark.asyncio
 async def test_small_cross_evaluation_gen7():
     await asyncio.wait_for(
-        simple_cross_evaluation(5, format_="gen7randomdoublesbattle"), timeout=10
+        simple_cross_evaluation(5, format_="gen7randomdoublesbattle"), timeout=20
     )
 
 
 @pytest.mark.asyncio
 async def test_small_cross_evaluation_gen8():
     await asyncio.wait_for(
-        simple_cross_evaluation(5, format_="gen8randomdoublesbattle", i=3), timeout=10
+        simple_cross_evaluation(5, format_="gen8randomdoublesbattle", i=3), timeout=20
     )
--- a/integration_tests/test_double_battles.py
+++ b/integration_tests/test_double_battles.py
@@ -35,12 +35,12 @@ async def simple_cross_evaluation(n_battles, format_, i=0):
 @pytest.mark.asyncio
 async def test_small_cross_evaluation_gen7():
     await asyncio.wait_for(
-        simple_cross_evaluation(5, format_="gen7randomdoublesbattle"), timeout=20
+        simple_cross_evaluation(3, format_="gen7randomdoublesbattle"), timeout=20
     )
 
 
 @pytest.mark.asyncio
 async def test_small_cross_evaluation_gen8():
     await asyncio.wait_for(
-        simple_cross_evaluation(5, format_="gen8randomdoublesbattle", i=3), timeout=20
+        simple_cross_evaluation(3, format_="gen8randomdoublesbattle", i=3), timeout=20
     )
--- a/integration_tests/test_simple_cross_evaluation.py
+++ b/integration_tests/test_simple_cross_evaluation.py
@@ -28,60 +28,61 @@ async def simple_cross_evaluation(n_battles, format_):
     await cross_evaluate(players, n_challenges=n_battles)
 
     for player in players:
+        player.reset_battles()
         await player.stop_listening()
 
 
 @pytest.mark.asyncio
 async def test_small_cross_evaluation_gen8():
     await asyncio.wait_for(
-        simple_cross_evaluation(5, format_="gen8randombattle"), timeout=10
+        simple_cross_evaluation(3, format_="gen8randombattle"), timeout=20
     )
 
 
 @pytest.mark.asyncio
 async def test_small_cross_evaluation_gen7():
     await asyncio.wait_for(
-        simple_cross_evaluation(5, format_="gen7randombattle"), timeout=10
+        simple_cross_evaluation(3, format_="gen7randombattle"), timeout=20
     )
 
 
 @pytest.mark.asyncio
 async def test_small_cross_evaluation_gen6():
     await asyncio.wait_for(
-        simple_cross_evaluation(5, format_="gen6randombattle"), timeout=10
+        simple_cross_evaluation(3, format_="gen6randombattle"), timeout=20
     )
 
 
 @pytest.mark.asyncio
 async def test_small_cross_evaluation_gen5():
     await asyncio.wait_for(
-        simple_cross_evaluation(5, format_="gen5randombattle"), timeout=10
+        simple_cross_evaluation(3, format_="gen5randombattle"), timeout=20
     )
 
 
 @pytest.mark.asyncio
 async def test_small_cross_evaluation_gen4():
     await asyncio.wait_for(
-        simple_cross_evaluation(5, format_="gen4randombattle"), timeout=10
+        simple_cross_evaluation(3, format_="gen4randombattle"), timeout=20
     )
 
 
 @pytest.mark.asyncio
 async def test_small_cross_evaluation_gen3():
     await asyncio.wait_for(
-        simple_cross_evaluation(5, format_="gen3randombattle"), timeout=10
+        simple_cross_evaluation(3, format_="gen3randombattle"), timeout=20
     )
 
 
 @pytest.mark.asyncio
 async def test_small_cross_evaluation_gen2():
     await asyncio.wait_for(
-        simple_cross_evaluation(5, format_="gen2randombattle"), timeout=10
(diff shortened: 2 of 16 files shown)


hsahovic:
@MatteoH2O1999 you can override the existing `EnvPlayer` class if you go in that direction!
Also, in case you missed it, you might be interested in setting `start_listening` to false and managing the event loop directly
Let me know if you want to discuss this!

MatteoH2O1999:
For an example check the relative example file:
https://github.com/hsahovic/poke-env/blob/0d4dc75f0772e0bc288e65ecbc7063ca502c41ce/examples/openai_example.py

hsahovic:
@MatteoH2O1999 thanks for this - i'll review it in depth over the weekend

integration_tests/test_simple_cross_evaluation.py
```diff
@@ -28,60 +28,61 @@ async def simple_cross_evaluation(n_battles, format_):
...
     for player in players:
+        player.reset_battles()
         await player.stop_listening()
 
 
 @pytest.mark.asyncio
 async def test_small_cross_evaluation_gen8():
     await asyncio.wait_for(
-        simple_cross_evaluation(5, format_="gen8randombattle"), timeout=10
+        simple_cross_evaluation(3, format_="gen8randombattle"), timeout=20
```
hsahovic, line 38: Did you encounter issues regarding execution time with this (and following tests)?

integration_tests/test_double_battles.py
```diff
@@ -28,18 +28,19 @@ async def simple_cross_evaluation(n_battles, format_, i=0):
...
     for player in players:
+        player.reset_battles()
         await player.stop_listening()
 
 
 @pytest.mark.asyncio
 async def test_small_cross_evaluation_gen7():
     await asyncio.wait_for(
-        simple_cross_evaluation(5, format_="gen7randomdoublesbattle"), timeout=10
+        simple_cross_evaluation(5, format_="gen7randomdoublesbattle"), timeout=20
```
hsahovic, line 38: Did you encounter issues regarding execution time with this (and following tests)?

src/poke_env/player/env_player.py
```diff
@@ -76,127 +67,23 @@ def __init__(
             Defaults to None.
         :type team: str or Teambuilder, optional
         """
-        super(EnvPlayer, self).__init__(
+        self._reward_buffer = {}
+        self.opponent_lock = Lock()
+        self.opponent: Optional[Union[Player, str]] = opponent
```
hsahovic, line 72: If no opponent is passed, we should default to a `RandomPlayer` instance playing the same format.

src/poke_env/player/openai_api.py
```diff
@@ -0,0 +1,537 @@
...
+        )
+        task.result()
+
+    async def async_join(self):
+        await self.queue.join()
+
+
+class _AsyncPlayer(Player):
+    def __init__(self, user_funcs, username, **kwargs):
+        self.__class__.__name__ = username
```
hsahovic, line 122: What is this for?

src/poke_env/player/openai_api.py
```diff
@@ -0,0 +1,537 @@
...
+        self.action_space = Discrete(self.action_space_size())
+        self.observation_space = self.describe_embedding()
+        self.current_battle: Optional[AbstractBattle] = None
+        self.last_battle: Optional[AbstractBattle] = None
+        self._keep_challenging: bool = False
+        self.challenge_task = None
+        if start_challenging:
+            self._keep_challenging = True
+            self.challenge_task = asyncio.run_coroutine_threadsafe(
+                self.challenge_loop(), asyncio.get_event_loop()
```
hsahovic, line 216: Would it be possible to:
- test what the current event loop is, and if necessary change it to `THREAD_LOOP` and
- then pass `THREAD_LOOP` directly, and therefore avoid the context manager?

hsahovic (changes requested):
This is great work - thanks for taking the time to deep dive in it and putting the work.

The general idea is better than the option I initially had in mind to achieve this. It's simpler and more elegant. I think we should move forward with it.

I added a couple of simple comments, but I think that the main points that need refinement are mainly linked to better default and, if possible, a simpler API.

Two additional things:
- is there a need to keep `EnvPlayer` and `OpenAIGymEnv` two separate entities?
- could you add an example using some RL framework?

I will do a deeper review later - if there is any specific issue you'd like my input on, let me know either here or on discord :)

integration_tests/test_simple_cross_evaluation.py
MatteoH2O1999: The timeout is to counter cases where battles happen to be very long. It didn't happen all that often but I think it's better to be on the safe side with this. This way you know if a test fails, it's not because all the battles took 100 turns.
Regarding the number of battles it's a matter of memory constraint on CircleCI: a medium docker container allows 4 Gb or RAM. The server alone takes up 1.5 Gb and the tests are all called with a single pytest (which brings its own overhead) and, to my knowledge, without garbage collecting the battles. During testing i noticed a lot of times CircleCI killed the build because of RAM constraints.

src/poke_env/player/env_player.py
MatteoH2O1999: I agree, will push that on the next batch

src/poke_env/player/openai_api.py
MatteoH2O1999: This avoids all the players to use as username "_AsyncPlayer xyz". It instead uses the name of the class that contains the player

src/poke_env/player/openai_api.py
MatteoH2O1999: The problem with this is that it wouldn't work: for this to work every single agent that will interact with the `OpenAIGymEnv` HAS to be listening on the `THREAD_LOOP`. This can be achieved by what you proposed plus writing in the docs that every agent that needs to interact with the `OpenAIGymEnv`  needs to be creted with `start_listening=False` and then to have it listen on the correct loop. I think the context manager simplifies this and allows to be sure that when outside of the context the main event loop (if present) is used

src/poke_env/player/openai_api.py
MatteoH2O1999: It is a bit hacky, I know ;)

MatteoH2O1999:
> * is there a need to keep `EnvPlayer` and `OpenAIGymEnv` two separate entities?

To be honest, I would prefer to keep them separate as `OpenAIGymEnv` is the implementation of the Gym API and allows more freedom for creating complex classes, while `EnvPlayer` is something easy to use out-of-the-box and used as the base for the single battle env player implementations with pre-implemented `action_to_move` methods, so the user would only need to implement `calc_reward`, `embed_battle` and `describe_embedding`. In short, you have one class implementing the Open AI API (`OpenAIGymEnv`) and one default implementation (`EnvPlayer`). This is the reason I would keep them as separated files.
> * could you add an example using some RL framework?

Sure, but I would prefer to wait until we settled on the API, just to avoid continuously rewriting it if we decide to change something. On that matter, could you explain a little more what you meant by "simpler API" (I'm genuinely curious as I'm only 22 and this is my first "real" experience on something a little bigger than a uni project)? Like I honestly don't find it all that complex as a lot of it is for more advanced work and if you set `start_challenging=True` you can literally only use the OpenAI methods (`reset`, `step`, `seed`, `render` and `close`) (once again in my very unexperienced opinion)

src/poke_env/player/openai_api.py
hsahovic: Oh, got it. Definitely hacky, but that's ok :)

src/poke_env/player/openai_api.py
hsahovic: That makes sense. Let me think about it a bit more - I'll get back to you tomorrow

MatteoH2O1999: Hey,
I get where you're going with this and I think that while it's possible to implement it wouldn't remove the context manager: when you need to call an async method of the player you would need to call it using:
```
task = asyncio.run_coroutine_threadsafe(player.battle_against(opponent, 1), player.loop)
```
Same thing for the unified thread:
```
task = asyncio.run_coroutine_threadsafe(player.battle_against(opponent, 1), BACKGROUND_LOOP)
```
And then collect the result using:
```
task.result()
```
While this is true also for the context manager:
```
task = asyncio.run_coroutine_threadsafe(player.battle_against(opponent, 1), asyncio.get_event_loop())
```
the context manager would allow users to call other players normally with `await` when outside of it, so it would retain backward compatibility.

MatteoH2O1999: I think the best solution is the shared background thread then. I'll start working on it tomorrow

MatteoH2O1999: Hey @hsahovic,
I opened a new draft with the changes you requested but could you consider merging #264?
I explained the rationale for it in the PR

Thank you

hsahovic: I merged #264 but I worry medium+ might not actually be enough - let me know if you think we should go another step up in RAM

MatteoH2O1999: Hey @hsahovic,
take a look at #263.
It should do what you wanted without actually changing the API :)

Let me know if the API are ok so I can get on writing the tutorials and the documentation

Issue #251, opened 2022-01-18
Pokemon volatileStatus


VolundrAI:
This might be a simple issue of not finding where to check, but some moves like minimize cause a "volatileStatus" that doesn't seem to show up in effects, side conditions, or anywhere that I can find.

Is this something that still needs to be implemented, or is there somewhere we can check that a Pokemon is impacted by a specific volatile status.

Thanks in advance!

VolundrAI:
I created a pull request solving this issue if you want to take a look at it

cameronangliss:
#261 and #556 seem to address this issue, closing now.

PR #261, opened 2022-02-10
Added support for Minimize as an Effect

VolundrAI:
Minimize is a silent effect that causes increased damage from certain moves https://bulbapedia.bulbagarden.net/wiki/Minimize_(move)

I opened an issue https://github.com/hsahovic/poke-env/issues/251 asking about this feature, but went ahead and added support for this specific move myself.

Let me know if there are any changes that should be made to this commit to get it approved


hsahovic:
Thanks @VolundrAI !

PR #272, opened 2022-03-25
Add data update workflow

MatteoH2O1999:
Hey @hsahovic 

This will allow to keep the data up-to-date with the showdown server. It will create (or update) a pull request with the updated data every day at midnight (dw, actions are completely free for public repos). Let me know what you think :)


hsahovic:
@MatteoH2O1999 this is a great idea - thanks a lot!

PR #294, opened 2022-06-13
Fix inconsistency between values returned by won() and lost()

### LINKED ISSUE #293: Inconsistency between battle.won and battle.lost
`abstract_battle.py`

```
@property
    def lost(self) -> Optional[bool]:
        """
        :return: If the battle is finished, a boolean indicating whether the battle is
            lost. Otherwise None.
        :rtype: Optional[bool]
        """
        return self._won is False

    @property
    def won(self) -> Optional[bool]:
        """
        :return: If the battle is finished, a boolean indicating whether the battle is
            won. Otherwise None.
        :rtype: Optional[bool]
        """
        return self._won
```

`won()` can return any of (`None`, `True`, `False`) but `lost()` will always return one of (`True`, `False`) since `None is False`=True

akashsara:
Fix #293

### DIFF
--- a/src/poke_env/environment/abstract_battle.py
+++ b/src/poke_env/environment/abstract_battle.py
@@ -872,7 +872,7 @@ class AbstractBattle(ABC):
             lost. Otherwise None.
         :rtype: Optional[bool]
         """
-        return self._won is False
+        return None if self._won is None else not self._won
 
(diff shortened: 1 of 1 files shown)


hsahovic:
Thanks @akashsara !

PR #300, opened 2022-06-29
Add assertion messages

MatteoH2O1999:
Hey @hsahovic,
I added messages to all the assertions in the code so future bugs will be easier to identify.

Let me know what you think and if you want to change something


hsahovic:
This is a great idea, thanks @MatteoH2O1999 !

PR #304, opened 2022-07-18
Add ping interval and timeout to player

MatteoH2O1999:
Sometimes showdown server could suffer from a lag spike, expecially on less powerful machines. This allows to increase or disable timeouts for the keepalive mechanism used by websockets.

@hsahovic let me know if you think this could be useful

### DIFF
--- a/src/poke_env/player/env_player.py
+++ b/src/poke_env/player/env_player.py
@@ -34,6 +34,8 @@ class EnvPlayer(OpenAIGymEnv, ABC):
         server_configuration: Optional[ServerConfiguration] = None,
         start_listening: bool = True,
         start_timer_on_battle_start: bool = False,
+        ping_interval: Optional[float] = 20.0,
+        ping_timeout: Optional[float] = 20.0,
         team: Optional[Union[str, Teambuilder]] = None,
         start_challenging: bool = True,
     ):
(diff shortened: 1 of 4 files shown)


MatteoH2O1999:
Added a fix for metaclass conflict as reported in #305

hsahovic:
@MatteoH2O1999 what's the default timeout ping? How did you encounter this issue?

SomeRandomGuy009:
> Added a fix for metaclass conflict as reported in #305

yep, this does fix the issue.

MatteoH2O1999:
@hsahovic,
default timeout is 20 seconds. For training I'm using an old computer so the server sometimes won't respond in time because of tensorflow keeping all the CPU occupied or because of RAM constraints forcing a lot of page swapping

hsahovic:
@MatteoH2O1999 this looks good - can you remove the metaclass fix (I fixed the gym version instead)?

MatteoH2O1999:
Sure, but won't pinning the gym version be limiting?

hsahovic:
It is, but the new gym version made some deeper changes to their base API (eg. what `step` should be returning) and I'd prefer to wait before switching to it, so we can assess if it also requires other changes.

MatteoH2O1999:
Sure thing: I'll take a look at it and see if I can figure something out. In the meantime I already removed the metaclass fix

PR #308, opened 2022-07-28
Account for get_event_loop() deprecation warning in Python 3.10

MatteoH2O1999:
Account for future deprecation of get_event_loop method of asyncio. get_running_loop is threadsafe so the effect will be the same

### DIFF
--- a/src/poke_env/player/player_network_interface.py
+++ b/src/poke_env/player/player_network_interface.py
@@ -75,7 +75,16 @@ class PlayerNetwork(ABC):
 
     @staticmethod
     def _create_class(cls, *args, **kwargs):  # pragma: no cover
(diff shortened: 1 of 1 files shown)


hsahovic:
Thanks @MatteoH2O1999 ! I'm thinking of dropping 3.6 support with the next poke-env release and adding 3.10. I'll get back to this PR at that time.

MatteoH2O1999:
With this fix it should still be compatible with 3.6 but you're right, 3.6 is getting a little bit old ;)

hsahovic:
Thanks @MatteoH2O1999 ! I'm going to merge this one and then get rid of the exception catching since I dropped python 3.6 support :)

Issue #306, opened 2022-07-25
Error encountered during player.ladder()


akashsara:
Hi,
I was testing a model I trained on Pokemon Showdown (code snippet below) when I ran into this issue. I'm able to challenge the bot to a battle and play against it perfectly well but when I do `player.ladder(100)` it errors out after completing a single battle. 

```
2022-07-25 18:33:47,574 - UABGLSimpleDQN - ERROR - Unhandled exception raised while handling message:
>battle-gen8randombattle-1625188644
|-message|Nukkumatti lost due to inactivity.
|
|win|UABGLSimpleDQN
Traceback (most recent call last):
  File "E:\Dev\meta-discovery\torch_env\lib\site-packages\poke_env\player\player_network_interface.py", line 131, in _handle_message
    await self._handle_battle_message(split_messages)
  File "E:\Dev\meta-discovery\torch_env\lib\site-packages\poke_env\player\player.py", line 235, in _handle_battle_message
    self._battle_finished_callback(battle)
  File "E:\Dev\meta-discovery\torch_env\lib\site-packages\poke_env\player\env_player.py", line 106, in _battle_finished_callback
    self._observations[battle].put(self.embed_battle(battle))
KeyError: <poke_env.environment.battle.Gen8Battle object at 0x000001E1988D2EA0>
Task exception was never retrieved
future: <Task finished name='Task-39' coro=<PlayerNetwork._handle_message() done, defined at E:\Dev\meta-discovery\torch_env\lib\site-packages\poke_env\player\player_network_interface.py:117> exception=KeyError(<poke_env.environment.battle.Gen8Battle object at 0x000001E1988D2EA0>)>
Traceback (most recent call last):
  File "E:\Dev\meta-discovery\torch_env\lib\site-packages\poke_env\player\player_network_interface.py", line 177, in _handle_message
    raise exception
  File "E:\Dev\meta-discovery\torch_env\lib\site-packages\poke_env\player\player_network_interface.py", line 131, in _handle_message
    await self._handle_battle_message(split_messages)
  File "E:\Dev\meta-discovery\torch_env\lib\site-packages\poke_env\player\player.py", line 235, in _handle_battle_message
    self._battle_finished_callback(battle)
  File "E:\Dev\meta-discovery\torch_env\lib\site-packages\poke_env\player\env_player.py", line 106, in _battle_finished_callback
    self._observations[battle].put(self.embed_battle(battle))
[...]

PR #311, opened 2022-08-02
Gym new step API update

MatteoH2O1999:
Hey @hsahovic,
I updated the openAI wrapper to use the new API. Right now it is disabled by default, I would wait for ML libraries to implement the changes before changing the default.

I also removed the default opponent in EnvPlayer as discussed in #306

Let me know what you think

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-gym==0.24.1
+gym
 numpy
 orjson
 requests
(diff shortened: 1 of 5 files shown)


hsahovic:
Thanks @MatteoH2O1999 - I'll take a look at it and the gym changes

MatteoH2O1999:
Added background methods for accepting and sending challenges with OpenAIGym as requested in #314

hsahovic:
This is good overall - making all of the internals weakly private makes sense, the addition of `background_accept_challenge` is great and moving forward with the new API is nice. Thanks again for your amazing work @MatteoH2O1999 !

One question though - why get rid of the default for `opponent`?
Advanced users will look into how to customize the agent they use for training, while providing a default makes it simpler for new users to get started. I can be persuaded to remove the default, but I'm curious regarding your ideas on it.

hsahovic:
I think this is a good solution yeah, removing `None` will make things clearer

hsahovic (approved):

hsahovic:
Thanks @MatteoH2O1999 !

Issue #312, opened 2022-08-03
reset_env() fails if no battles have been played yet


akashsara:
Hi, I'm using the latest version (as of two days ago) of the code here.

I call `player.reset_env()` multiple times to evaluate my model against different opponents. However, if the player hasn't played any battles before this call, it fails giving the following error:

`'NoneType' object has no attribute 'finished'`

I dived into the code and the problem seems to be in L520 of `openai_api.py`:

```
async def _stop_challenge_loop(
        # removed code above
        if force:
            if not self.current_battle.finished:  # pyre-ignore <---- this line
                if not self.actions.empty():
        # removed code below
```

The fix is simple `if self.current_battle is not None and not self.current_battle.finished` and I'm happy to put up a PR, but just wanted to check if this is indeed a bug and not intentional.

hsahovic:
I think this is a bug. I'd be happy to merge a PR for it :)

MatteoH2O1999:
Yep,
didn't notice that.
Thanks @akashsara !!!

hsahovic:
Thanks @akashsara ! I merged your fix in :)

PR #313, opened 2022-08-03
fix _stop_challenge_loop bug when no battles played

akashsara:
Solves #312.


hsahovic:
Thanks @akashsara ! For next time, don't hesitate to take a look at the cause of failing tests :)

PR #346, opened 2022-12-27
Create data submodule and reorganize static files

hsahovic:


hsahovic:
Side objectives:
- [ ] Nuke `Battle.from_format`

PR #347, opened 2022-12-29
Optimizing action CI pipeline

MatteoH2O1999:
Caching python dependencies does not actually give that much of an improvement and using venv instead of the base python installation is an additional unneded complexity, as setup-python already provides a clean python installation

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,3 +22,15 @@ updates:
   - dependency-name: pyre-check
     versions:
     - 0.0.59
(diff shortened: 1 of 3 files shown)


MatteoH2O1999:
Hey @hsahovic,
while updating to github actions you forgot to activate them also for pull requests. This PR should fix that in addition to optimizing the pipeline to reduce the cached resources (as there is a limit) and complexity without losing speed or functionality.

hsahovic:
Thanks @MatteoH2O1999, this is indeed better

PR #351, opened 2023-01-22
Add basic gen 9 support

hsahovic:


kodecreer:
I had to do a lot more. The biggest pain in Gen 9 support is Revival blessing which is still causing me issues here and there annoyingly

PR #334, opened 2022-09-08
Fix breaking changes in OpenAI gym regarding step and reset return signatures

MatteoH2O1999:
Should fix #333

PR #359, opened 2023-02-20
Fix breaking changes in OpenAI gym regarding step and reset return signatures

MatteoH2O1999:
Follows from #334 to fix merge conflicts without rebasing as the old branch is useful for current projects using the fix


MatteoH2O1999:
@hsahovic Now that integration tests work this can get merged in order to ensure compatibility both with legacy as well as new OpenAI Gym API (as well as Gymnasium as long as they don't change it again)

PR #360, opened 2023-02-21
Ensure examples run successfully in CI test workflow

MatteoH2O1999:
With this PR, all examples are checked in order to ensure they run successfully


hsahovic:
@MatteoH2O1999 this is a great idea, but I'm not sure it should run with regular tests; tests are already slow enough not to add entire examples on top of them.
What do you think of a weekly run, independently from the main test suite?

PR #387, opened 2023-06-21
Gen 9 support. Missing things include

kodecreer:
Missing
- Support for Revival blessing which for some reason doesn't update the status of the pokemon
- Random Player not updated to support revival blessing or Tera typing. Things not missing (Tested on 100k+ battles)

Includes
- Ability to terastialize and record teratialize typing
- Newest Pokemon from the soon to be DLC like Walking Wake etc...
- Newest moves included
- Gen 9 abilites and battle field effects included


kodecreer:
I would suggest to change typing for Terastializing in main class besides that. This works and I tested it on 100k+ battles. Issues that are around though include pokemon not being able to detect the change in revival blessing which is an issue....

PR #398, opened 2023-07-18
Move from `setup.py` to `pyproject.toml`

cameronangliss:
This PR brings `poke-env` from using `setup.py`  to using `pyproject.toml`, which is the modern way to describe how package installation proceeds. It's generally easier to read, removes the need for `MANIFEST.in`, and primes us in the future for configuring tools such as isort and passing stub files to allow `poke-env` importers to have all the type information from all imported classes/functions (when we get around to fully typing all of those out).

I also added the `flake8` `line-length` limit to the `black` configuration in the `pyproject.toml` file.

I've tested from this branch locally, and poke-env `pip install`s perfectly for me.


cameronangliss:
Currently, only a few features from `setup.py` are not brought into `pyproject.toml`. This includes:

1. `long_description` is omitted.
2. Processes to handle `version` not being specified are not included.
3. `UploadCommand` is omitted.

In the case of 1), I wasn't sure that it was necessary due to the new `readme` option added under `[project]` in `pyproject.toml`. As for 2), I didn't foresee `version` ever being unspecified. Finally, for 3), I spent some time trying to figure out how to get that in `pyproject.toml`, but I couldn't find a good solution, so I decided to simply remove it in the hope that maybe we don't really need that anymore. I believe its main functionality was to clean out the `dist` folder for every build of the package, but I think that may be the default behavior (correct me if I'm wrong).

@hsahovic - if you think any of the changes are undesirable, I'd love to discuss with you. I'm assigning this over to you now.

hsahovic:
Thanks @cameronangliss! I'll run some tests with this new setup this week and let you know if anything needs to be changed.

hsahovic:
Thanks @cameronangliss !

PR #441, opened 2023-10-29
feat: enable custom replay folder

tonysun9:
Enables a custom replay folder


hsahovic (changes requested):
This is already possible with the existing`save_replays` arg: if it's a string, it is used as the folder in which replays are saved.
I'm ok with with adding a separate `replay_folder` for clarity, but in that case please remove the existing string-based functionality; this is the kind of thing I'd prefer a unique way to accomplish.

tonysun9:
> This is already possible with the existing`save_replays` arg: if it's a string, it is used as the folder in which replays are saved. I'm ok with with adding a separate `replay_folder` for clarity, but in that case please remove the existing string-based functionality; this is the kind of thing I'd prefer a unique way to accomplish.

Thanks, I didn't realize this. Let me take a look.

tonysun9:
After looking at the code, I don't think this provides a meaningful benefit. Will look to improve things elsewhere.

Issue #382, opened 2023-06-12
check_env problem 


KlausRyu:
I was start to try debug the reinforcement learning example, but i get this error, is it gym problem or it is the code bug.

![Capture1](https://github.com/hsahovic/poke-env/assets/58900577/b7f1c041-9d6d-4570-ae75-a983b9922f7d)

KlausRyu:
solved by downgrade from 0.26 to <0.26

KlausRyu:
and i get error.
![Capture2](https://github.com/hsahovic/poke-env/assets/58900577/0362ccfb-5c10-4467-b2c6-4d87e5a8f684)

cameronangliss:
@hsahovic I believe this issue is resolved, do you agree?

hsahovic:
Yes - good catch @cameronangliss

PR #472, opened 2023-12-17
Complying with `check_env`

cameronangliss:
Addressing the warnings that popped up in #382.

The `seed` function is deprecated and `return_info` as an argument to the `reset` function is also deprecated, so we remove these.

I checked with an environment of my own, and these changes did indeed remove those warnings. The only warning left speaks to the fact that we don't use `gymnasium.make` to create our environments, but this isn't really feasible since each user makes their own custom environment that inherits from one of the base environments provided in poke-env.


hsahovic:
Thanks for your great work @cameronangliss !

Issue #525, opened 2024-04-03
Adding enumerations for various conditions


caymansimpson:
I think it would be great to add full enumerations of volatile statues and target types so that we can better OHE moves/pokemon states for neural net inputs; these are already supported for other enums

hsahovic:
Yeah, that makes sense!

Issue #523, opened 2024-04-03
Converting showdown targets into array indices


caymansimpson:
Right now, it's very hard to specify what mon should be targeting what other mon in double, since showdown targets don't map cleanly to arrays. I had to do a workaround mapping function solution in a forked version of poke-env here: https://github.com/caymansimpson/poke-env/blob/5f09a6d814b294c0c6827131485101afa2f64668/src/poke_env/utils.py#L90

I think it would be great to have "/move hyperbeam 0" to target the mon in the 0th position of the opponent's team's tuple

hsahovic:
Yeah, this is pretty bad in terms of usability.
What do you think would be the clearest API? Imo, it would be ideal to abstract away positions, for instance by intelligently habndling the `move_target` arg in `BattleOrder`. Does that make sense?

caymansimpson:
I think that makes sense -- so people will just use the index in provided in poke-env, and we will handle everything else? If so, how do we handle self-targets? (when battle.opponent_active_pokemon is 0 or 1, mapping to 1 and 2, and self-targets are -1 or -2)

Maybe enums would be best?

caymansimpson:
Combining with: https://github.com/hsahovic/poke-env/pull/529

PR #529, opened 2024-04-06
Adding enumerations for the target property of moves

caymansimpson:
Adding enumerations for the .target() field in Move class, to make it easier to embed moves via OHE.

Changes:
1. Created an enumeration class of Target
2. Changed move.target to return this enumeration
3. Added unit tests to test the enumeration
4. Added unit tests to test target in moves

All unit tests passed in both test_move and test_enumeration

This is in response to [issue #525](https://github.com/hsahovic/poke-env/issues/525), but only addresses the Target field because you already have encoded volatile statuses in effect.py

### DIFF
--- a/src/poke_env/environment/target.py
+++ b/src/poke_env/environment/target.py
@@ -8,7 +8,6 @@ import logging
 # from poke_env.environment.pokemon import Pokemon
 import re
 from enum import Enum, unique, auto
-from typing import Optional
 
 
 # This is an enum for all the targets you can have
--- a/src/poke_env/environment/target.py
+++ b/src/poke_env/environment/target.py
@@ -3,7 +3,12 @@ targets a move can have
 """
 
 import logging
+# from poke_env.environment.double_battle import DoubleBattle
+# from poke_env.environment.pokemon import Pokemon
+import re
 from enum import Enum, unique, auto
+from typing import Optional
+
 
 # This is an enum for all the targets you can have
 @unique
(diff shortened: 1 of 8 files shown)


caymansimpson:
Please reject if you think this is not well-thought through. I'm not sure how this should best jive with [Issue #523](https://github.com/hsahovic/poke-env/issues/523), and specifically the DoubleBattle implementation [here](https://github.com/hsahovic/poke-env/blob/c352e3590a44f9919374fd95f9b1e643e2fd031c/src/poke_env/environment/double_battle.py#L12-L16) and [here](https://github.com/hsahovic/poke-env/blob/c352e3590a44f9919374fd95f9b1e643e2fd031c/src/poke_env/environment/double_battle.py#L247-L330)

src/poke_env/environment/target.py
```diff
@@ -0,0 +1,57 @@
...
+import logging
+from enum import Enum, unique, auto
+
+# This is an enum for all the targets you can have
+@unique
+class Target(Enum):
+    """Enumeration, representing targets for each move
+    in a battle."""
+
+    UNKNOWN = auto()
```
hsahovic, line 14: do we expect any unknown value? if not, it might be better to raise an exception

src/poke_env/environment/target.py
```diff
@@ -0,0 +1,57 @@
+"""This module defines the Target class, which represents the types of
+targets a move can have
+"""
+
+import logging
+from enum import Enum, unique, auto
+
+# This is an enum for all the targets you can have
+@unique
+class Target(Enum):
```
hsahovic, line 10: i think i'd prefer `ADJACENT_ALLY` style names to `ADJACENTALLY` - they are a but more readable

hsahovic, line 10: we probably want a method that converts (self, pokemon, battle) to a valid showdown target

src/poke_env/environment/target.py
caymansimpson: This is a good callout. Will change!

src/poke_env/environment/target.py
caymansimpson: Sounds good - will convert!

caymansimpson: Great; I'll combine this with [Issue #523](https://github.com/hsahovic/poke-env/issues/523) then. 

I'll implement: `def to_showdown_target(self, order: BattleOrder, battle: AbstractBattle) -> int:`

It assumes in that DoubleBattleOrder(first_order, second_order) will map to first_order maps to the pokemon at 0th index in battle.active_pokemon and second_order maps to the pokemon at the 1st index in battle.active_pokemon

caymansimpson: I encountered a problem -- we'll have circular imports if we try to import Pokemon and Double Battle into the Target class, since Target is needed in Player, which imports Battle, which imports Move, which imports Target.

To remedy, I'll write this function in the DoubleBattle class as:
`def get_showdown_target(self, move: Move, target_mon: Pokemon) -> Optional[int]:`

(DoubleBattle, since targets are not needed for SingleBattle)

hsahovic:
@caymansimpson thanks!

PR #531, opened 2024-04-08
Add DoublesMaxBasePowerPlayer

AFumis:
Adds DoublesMaxBasePowerPlayer, a simple rule-based agent for double battles that picks the move with the highest BP, taking into account moves that hit both targets and the spread damage modifier. If the chosen move is single target, it picks randomly between opponents (but not the ally).


src/poke_env/player/doubles_baselines.py
```diff
@@ -0,0 +1,49 @@
...
+        if battle.available_moves:
+            my_order = []
+            for i in slots:
+                if battle.active_pokemon[i] and battle.available_moves[i]:
+
+                    # picks move by highest BP, takes into account spread damage modifier and hitting both targets
+                    best_move_lambda_key = lambda move: (
+                        move.base_power
+                        if move.target == "normal"
+                        else move.base_power * 0.75 * 2
```
hsahovic, line 24: What do you think of counting how many opponents are alive?

hsahovic, line 24: I think this should take into account how many non-fainted pokemons are active on the opponent's side.

src/poke_env/player/doubles_baselines.py
```diff
@@ -0,0 +1,49 @@
...
+                    )
+                    best_move_ = max(
+                        battle.available_moves[i], key=best_move_lambda_key
+                    )
+
+                    # randomly picks between the two opponents for normal move targeting
+                    targets = battle.get_possible_showdown_targets(
+                        best_move_, battle.active_pokemon[i]
+                    )
+                    opp_targets = [t for t in targets if t > 0]
```
hsahovic, line 34: `t > 0` is a bit implicit - can you do an explicit check based on `OPPONENT_1_POSITION` / `OPPONENT_2_POSITION`?

src/poke_env/player/doubles_baselines.py
```diff
@@ -0,0 +1,49 @@
...
+                    if opp_targets:
+                        target = opp_targets[int(random.random() * len(opp_targets))]
+                    else:
+                        target = targets[int(random.random() * len(targets))]
+
+                    my_order.append([BattleOrder(best_move_, move_target=target)])
+
+                else:
+                    my_order.append([])
```
hsahovic, line 43: Shouldn't this be a random switch?

unit_tests/player/test_doubles_baselines.py
```diff
@@ -0,0 +1,73 @@
...
+        username="placeholder_username",
+        logger=None,
+        gen=8,
+    )
+    battle.player_role = "p1"
+    active_pikachu = Pokemon(species="pikachu", gen=8)
+    active_pikachu.switch_in()
+    battle._active_pokemon["p1a"] = active_pikachu
+
+    # player_pkg.Battle = PseudoBattle
```
hsahovic, line 22: i think this comment should be deleted

hsahovic (changes requested):
This is looking pretty good and almost ready to merge.
Can you:
- make sure tests are passing
- implement this logic in `MaxBasePowerPlayer` + rename the existing `choose_move` to `choose_singles_move` and update `choose_move` to call either singles or doubles depending on `player.format_is_doubles`?

hsahovic:
Following up on our discussion on discord, I forked your branch in: https://github.com/hsahovic/poke-env/pull/537
Considering this, I'm closing this PR. Thanks for your contribution!

PR #543, opened 2024-05-03
Created Observations class

caymansimpson:
Created an observations class. This PR:
1. Adds modular event-level tracking via Observation class
2. Stores turn-by-turn information
3. Adds unit tests


caymansimpson:
Ah - ef word. I messed up git and had an old commit behind my observations class 🤦🏽‍♂️ Killing this and will try to fix

PR #551, opened 2024-05-26
Add VolatileStatus

caymansimpson:
Added an Enum of VolatileStatus, plus some tests around it. This is to help with embedding moves


caymansimpson:
Ah shiet - added too many of my previous commits 🤦🏽‍♂️ Will rebase

PR #552, opened 2024-05-27
Adding Volatile Status

caymansimpson:
I wanted to add an enum for VolatileStatus to make it easier to embed. What this PR has:
1. Adding VolatileStatus (a semi-common field in moves data) as an enum
2. Changing Moves to return a VolatileStatus object, when requesting a move's volatile_status
3. Adding a unit test to moves to test VolatileStatus


caymansimpson:
Closing because @hsahovic recommended we merge this with effect

PR #556, opened 2024-06-02
Adding VolatileStatus

caymansimpson:
Added all VolatileStatus to Effects, added tracking in Effect for VolatileStatuses, whether effects come from abilities or moves, edited volatile_status in Move to return an Effect, added a way to convert strings in static data to Effects, and added testing

Idea here is that we are able to allow AI to better understand effects, and link them to moves/abilities


src/poke_env/environment/effect.py
```diff
@@ -243,6 +265,31 @@ def from_showdown_message(message: str) -> Effect:
...
+        :rtype: Effect
+        """
+        message = message.replace("_", "")
+        message = message.replace(" ", "")
+        message = message.replace("-", "")
+        message = message.upper()
+        try:
+            return _FROM_DATA[message]
+        except KeyError:
+            logging.getLogger("poke-env").warning(
```
hsahovic, line 284: Can you add a test for this path? Eg. calling `from_data` on a non-existing effect

hsahovic (commented):
This looks good, but I'm concerned about maintaining all categories of effects up to date.
Can you add:
- a check that all effects is from either an ability, item or move, and only one? (if this is true?)
- all volatile effects appear in at least one move?
- all move's volatile effects are a) an effect and b) volatile?
- any other automated check you can think of

caymansimpson:
It was a good call to add these. I think what you came up with was pretty comprehensive and I definitely caught some stuff. This test setup should ensure that if new volatile_statuses are added, Effects will get flagged (as long as we update the move_generator to the most recent gen). Because you manually updated Effects with all the other showdown effects, I'm not sure how to ensure they get added; however, if they are added, the person who added them will have to all ensure that the class variables also get updated too

hsahovic:
That sounds good - thanks @caymansimpson !

PR #558, opened 2024-06-06
Added gen and format information to AbstractBattle

caymansimpson:
I did a couple of things here:
1. `_format` was an internal variable that wasn't used, and recorded the generation of the `AbstractBattle`, so I changed `_format` to hold the format of the battle, aligning with Showdown protocol via [the python equivalent to Showdown's parsing](https://github.com/pkmn/stats/blob/fc987d77c0163d6a9f6c4822ce0850c7fb95f829/logs/src/config.ts#L18)
2. Stored generational info (the int) in `_gen`
3. Added `gen` and `format` as properties of `AbstractBattle`
4. Added a `RunTime` check to ensure that `AbstractBattle` is initiated with the right `gen`
5. Added to `fixture_data` official [showdown logs](https://github.com/pkmn/stats/blob/main/stats/src/test/testdata/logs/gen6doublesou/log.1.json) for easier `parse_message` testing
6. Added testing for `battle.format` and `battle.gen`


hsahovic (approved):
Great change!

PR #565, opened 2024-06-16
Enhancing ObservedPokemon and Adding naming consistency to terrastilized

caymansimpson:
1. Adding other fields we need to track from a mon for accurate speed and damage calculation (eg we need to record its ability if someone attacked it with worry seed or skill swap)
2. Changed terrastilized to is_terrastilized to be consistent with is_dynamaxed
3. Added testing for both


hsahovic:
@caymansimpson can you fix the linting issue?

caymansimpson:
Ah sorry about that 🤦🏽‍♂️ just updated

hsahovic (approved):

PR #569, opened 2024-06-26
A whole bunch of useful improvements

cameronangliss:
This PR is a bag of miscellaneous improvements, including:
1. A new `Player` method called `battle_against_multi` which allows you to play one Player instance against a list of Player instances without having to play everyone against everyone as is done in `cross_evaluate`. Also, `cross_evaluate` is simplified by using `battle_against_multi` internally.
1. There is no longer a timeout for connecting the websocket in the Showdown client.


cameronangliss:
@hsahovic this is ready to go whenever you're ready to take it!

cameronangliss:
Just as a note, I could be doing `reset_battles` on self and opponent in the `_battle_against` method, and that's probably the thing that makes most sense given that the function now returns the win and lose rates, but I didn't want to make a compatibility issue with the previous functionality. @hsahovic let me know if you care about compatibility with this function or not, and if not I can put `reset_battles` in there. Currently it's in `_battle_against_multi`.

cameronangliss:
After using this diff myself for a little while, I decided that it would be better to just to reset_battles in the battle_against method. I just put that change in.

hsahovic:
I agree that the current way results are handled by player objects can be improved, but i don't think starting battles should erase the previous results. These two actions should be covered by different methods in the API.
I'm open to a proposal for a better structure around it, but i'm not sure this is it. Let's chat on discord?
Your PR seems to contain other changes - do you want to open separate PRs for them?

cameronangliss:
@hsahovic sorry, I've been MIA for a little while, but I've cleaned this PR up and it's ready for review :)

hsahovic:
Hey @cameronangliss,
I missed your comments - sorry for the delay. Thanks for this PR and good update.

 >   A new Player method called battle_against_multi which allows you to play one Player instance against a list of Player instances without having to play everyone against everyone as is done in cross_evaluate. Also, cross_evaluate is simplified by using battle_against_multi internally.

I don't think this is a good API - however, making `battle_against` accept multiple opponents (either in *args or with an iterable) would probably be a good change, if you want to do so.

>    Some methods of Player are made static that don't need the self value, allowing users of poke-env to use those methods in their own static methods.
>     accept_open_team_sheets is possible to turn on for OpenAIGymEnv instances as well as Player instances now.
> The team being used by a Player instance is now re-rolled at the beginning of every game played if a TeamBuilder is specified.
>    A bug in the choose_move method of SimpleHeuristicsPlayer was fixed where a decision wouldn't be made if the active pokemon on either side is None.

These are good changes, i'd be happy to merge them

>    There is no longer a timeout for connecting the websocket in the Showdown client.

I would be ok with an optional parameter with a reasonable default value that can be set to infinity with a 0 or None, but no timeout by default is a gotcha in my opinion.

If you want to update this PR for these changes I'd be happy to review it - otherwise if you prefer we can also do one PR per change.
Let me know what you think!

PR #577, opened 2024-07-14
Added empty event tracking

caymansimpson:
In Poke-env showdown protocol, empty events are used to track where we are in a turn state ([singles example](https://github.com/pkmn/stats/blob/main/anon/src/testdata/anon.json#L427), [doubles example](https://github.com/pkmn/stats/blob/main/stats/src/test/testdata/logs/gen6doublesou/log.1.json)), which are both useful to parsers and LLMs.

This change is quite small -- it allows for us to record empty events in observations.events so that any AI/model can better track what "chapter" of the turn we're in.


caymansimpson:
This is included in my other PR, so closing this for now

Issue #576, opened 2024-07-14
Raw Stats not accessible until a Request


caymansimpson:
A mon's stats are not accessible until a Request occurs, which makes it hard to make damage calcs without parsing and storing the team.

Would you be open to a PR that:
1. Tracks stats w/ an internal `Pokemon._stats` variable?
2. Updates the getter for `Pokemon.stats`, changing it from `Pokemon._last_request` to `_stats` variable?
3. Adds a setter for `Pokemon.stats` in `Pokemon`?
4. Updates internal `_stats` variable in `Pokemon.update_from_request`
5. In `player._create_battle`, calculates and sets the stats, with help from `stats.py`?

PR #581, opened 2024-07-30
Adding stats storage, name in Pokemon, and teampreview storage

caymansimpson:
This is building off of this [PR](https://github.com/hsahovic/poke-env/pull/577) and refers to [this issue](https://github.com/hsahovic/poke-env/issues/576). It does the following:
1. Fixes `stats` storage, such that:
    a. We record the stats of our own pokemon via teambuilder so that we have access to them before a request happens
    b. We record our HP stat in stats
    c. Add a `stats` setter to support future moves that affect stats (e.g. power swap, speed swap)
2. Add a `name` property to Pokemon to make finding pokemon identifiers easier, especially when we need to parse through showdown logs manually
3. Stores the player's teampreview pokemon in a teampreview property. This is important to track because opponent moves are going to be affected by this information (for VGC)

I'm submitting this PR just because I think these changes are valuable for other developers using poke-env, but don't feel pressure to accept them!


caymansimpson:
After sitting with this, I will probably add one more update to this to change teampreview_team to a List instead of a Set (and resolve the conflict). It shouldn't block this PR, but I think a List would help because in player.teampreview, having an ordered list of mons is more helpful, as we have to return the indices + 1 as a message

src/poke_env/player/player.py
```diff
@@ -213,6 +213,15 @@ async def _create_battle(self, split_message: List[str]) -> AbstractBattle:
                         save_replays=self._save_replays,
                     )
 
+                # Add our team as teampreview_team, as part of battle initialisation
+                if isinstance(self._team, ConstantTeambuilder):
```
hsahovic, line 217: is there data in here that we don't get during teampreview?

src/poke_env/environment/abstract_battle.py
```diff
@@ -1221,6 +1228,18 @@ def teampreview(self) -> bool:
         """
         return self._teampreview
 
+    @property
+    def teampreview_team(self) -> Set[Pokemon]:
```
hsahovic, line 1232: what is currently happening for this in teampreview? is our team not accessible in `self.team`? if so, why not store it in teampreview_team instead?

src/poke_env/player/player.py
caymansimpson: Yes! It is primarily HP stats and nature (which interacts with berries). Though now that I look again at the teampreview request, you can get HP stats from "condition", and I think berry interactions are probably too niche and rare. I'm thinking I should simplify and remove?

src/poke_env/environment/abstract_battle.py
caymansimpson: I think there are two philosophical reasons I think it would make sense to store teampreview_team. I say "philosophical" because I could see people falling on either side and being well-thought through.

Philosophical idea: The battle object is supposed to capture the state of the battle, as it pertains to making optimal decisions. This means it should capture not only what we know, but also what the opponent knows. Thus, we need to store what the opponent thinks we might have for when we pass the battle state into our Deep Learning network of choice (you can also make the case that this should be done via player, which would also be reasonable).

src/poke_env/player/player.py
hsahovic: i think it makes sense to store this somewhere, but i'm not sure this is the right way to do it. the main change i'd like is instead of relying of the specifics of the constant teambuilder, store the generated team for each battle and inject it in the battle on teampreview requests
that being said, this implementation is a good first step - maybe we keep this as a first version and if we want to improve how we deal with this in the future we refactor it? i think this would also cover most cases

src/poke_env/environment/abstract_battle.py
hsahovic: i see - this is reasonnable

hsahovic:
@caymansimpson can you take a look at the conflicts and then i'll merge?

caymansimpson:
Resolved!

Issue #579, opened 2024-07-23
Temporary Effects


caymansimpson:
There are temporary `Effect`s that only last a turn; for example:
- Quick Claw
- After You
- Custap Berry
- After You
- Flinch
- Helping Hand

But end silently in Showdown (eg they don't have a `['', '-end',...]` message). This means that once a `Pokemon` has the `Effect` applied to them, this effect incorrectly persists for the rest of the game in poke-env. Some end after a pokemon moves, and some end after the turn, and some end when switched out (among other reasons). Would you be open to a PR that adds logic for these various effects? I don't want AI that encodes the battle state to think that once you Helping Hand a mon, it will always be helping handed. It will also make things like damage calculations with poke-env very difficult

I'm not sure what the best way to maintain/test implementation logic is though, given that it is almost necessarily custom. Curious if you have thoughts?

caymansimpson:
@hsahovic I'm thinking of implementing by:
1. Adding a list of effects that go away after moves (e.g. glaiverush)
2. Adding a list of effects that go away after switches (e.g. all volatile statuses)
3. Adding a list of effects that go away after turns (e.g. After You)

I will add these as curated lists in effects.py and implement the checks in abstract_battle.py for the next PR. I'm not going to get 100% coverage of every effect, but I should be able to get at least 95%. LMK if you have any opinions on (1) implementation plan, (2) how I should test or (3) anything else!

PR #593, opened 2024-08-24
Added Effect Handling, and reordered a set for lint

caymansimpson:
[Issue here](https://github.com/hsahovic/poke-env/issues/579)

This just removes Effects when they silently get removed in the true battle state, but Showdown doesn't send you a log. Like all effects, testing is quite manual and maintenance is hard given that each effect is typically a unique edge-case.

However, this is 95% there and unblocks reverse speed/damage calculation

I also reordered items added to a set in abstract_battle to be consistent


hsahovic (approved):

hsahovic:
Thanks @caymansimpson !

Issue #628, opened 2024-11-13
Add Override handling for pledge moves


caymansimpson:
Pledge moves are incorrectly overrode and incorrectly assigned to the mon (e.g. the second mon to use a pledge move gets both pledge moves assigned to its moves property). This creates errors that crashes poke-env when a last_request has different moves than ones assigned to it.

Will adjust the handling in battle.parse_message

hsahovic:
Fixed by https://github.com/hsahovic/poke-env/pull/637 - thanks @caymansimpson !

PR #637, opened 2024-11-15
Added correct handling for pledge moves

caymansimpson:
Right now, pledge moves are incorrectly handled and overrode, so that if Piplup used Water Pledge and Cyndaquil followed up with Fire Pledge, we would set Cyndaquil's move as Water Pledge. This is a problem when we update the mon with a request, because there'd be a conflict of moves and it will raise a Error

Mentioned here: https://github.com/hsahovic/poke-env/issues/628

### DIFF
--- a/unit_tests/environment/test_double_battle.py
+++ b/unit_tests/environment/test_double_battle.py
@@ -307,3 +307,43 @@ def test_gen_and_format(example_doubles_logs):
     assert battle.gen == 6
     assert battle.battle_tag == "tag"
     assert battle.format == "gen6doublesou"
+
+
+def test_pledge_moves():
+    battle = DoubleBattle("tag", "username", MagicMock(), gen=8)
+    battle.player_role = "p2"
+
+    events = [
+        ["", "switch", "p1a: Indeedee", "Indeedee-F, L50, F", "100/100"],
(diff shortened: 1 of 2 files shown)


unit_tests/environment/test_double_battle.py
```diff
@@ -307,3 +307,43 @@ def test_gen_and_format(example_doubles_logs):
...
+    battle = DoubleBattle("tag", "username", MagicMock(), gen=8)
+    battle.player_role = "p2"
+
+    events = [
+        ["", "switch", "p1a: Indeedee", "Indeedee-F, L50, F", "100/100"],
+        ["", "switch", "p1b: Hatterene", "Hatterene, L50, F", "100/100"],
+        ["", "switch", "p2a: Primarina", "Primarina, L50, F, shiny", "169/169"],
+        ["", "switch", "p2b: Decidueye", "Decidueye-Hisui, L50, F, shiny", "171/171"],
+        ["", ""],
+        ["", "-terastallize", "p2a: Primarina", "Poison"],
```
hsahovic, line 322: why tera here? or is it just copied from a raw battle log?

caymansimpson: Ah yeah 😅 I just copied it from my fuzz testing logs. Ill edit it down for clarity!

hsahovic (approved):

Issue #625, opened 2024-11-13
Add name property to Pokemon


caymansimpson:
You can add nicknames to pokemon in-game (or in sowdown) which will become their showdown identifier. I am adding this as a field to the Pokemon class so that given a Pokemon, we can reconstruct its showdown identifier. This helps with battle parsing

hsahovic:
Fixed by https://github.com/hsahovic/poke-env/pull/638
Thanks @caymansimpson !

PR #638, opened 2024-11-15
Fixed name and teambuilder handling

caymansimpson:
Did two things:
1. Added `name` property to `Pokemon` so that it can store their own Showdown identifiers; it is an optional argument that can be passed, which is typically passed through a request
2. Fixed the reliability of Teambuilder instantiation for `Pokemon`, where a `nature` isn't needed to be specified. In the old setup, we wouldn't calculate the stats because we didn't have the nature. But in reality, Showdown assumes no nature is `"serious"`, and so we correct for this to ensure we have stats calculated for any mon that is bulit via teambuilder'

Added appropriate testing

Mentioned here: https://github.com/hsahovic/poke-env/issues/625

### DIFF
--- a/src/poke_env/environment/pokemon.py
+++ b/src/poke_env/environment/pokemon.py
@@ -33,6 +33,7 @@ class Pokemon:
         "_max_hp",
         "_moves",
         "_must_recharge",
+        "_name",
         "_possible_abilities",
         "_preparing_move",
         "_preparing_target",
@@ -55,6 +56,7 @@ class Pokemon:
         gen: int,
         *,
         species: Optional[str] = None,
+        name: Optional[str] = None,
         request_pokemon: Optional[Dict[str, Any]] = None,
         details: Optional[str] = None,
         teambuilder: Optional[TeambuilderPokemon] = None,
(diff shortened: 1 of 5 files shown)


src/poke_env/environment/pokemon.py
```diff
@@ -654,9 +668,7 @@ def base_species(self) -> str:
         :rtype: str
         """
         dex_entry = self._data.pokedex[self._species]
-        if "baseSpecies" in dex_entry:
-            return to_id_str(dex_entry["baseSpecies"])
-        return self._species
+        return to_id_str(dex_entry["baseSpecies"])
```
hsahovic, line 671: is this always defined? and why is this change in this PR?

caymansimpson: Ah, I'm sorry! This isn't strictly necessary for the PR -- would you like me to remove it?

For context, I think you already ensured that each species has a `baseSpecies` in `GenData` [here](https://github.com/hsahovic/poke-env/blob/a2cd6e8e5c88c0fac5f3f61d67d150af8a1253e4/src/poke_env/data/gen_data.py#L66) and so these lines were duplicative; I just cleaned this up for simplicity because it was hard for me to track down what baseSpecies actually was when I was doing this work

hsahovic: Ah, i see. Ok, this makes sense.

hsahovic (approved):

Issue #629, opened 2024-11-13
Add support for typechange and skillswap


caymansimpson:
Right now typechange and skillswap is a non-niche interaction in VGC that is not handleable in poke-env today. Planning to:
- modify the error of Effect.TYPE_CHANGE -> Effect.TYPECHANGE
- add support for typechange using a new temporary_types internal field
- add support for skill swap using a new temporary_ability field
- edit end_effect to also undo the temporary changes above
- change clear_effects to end effect each effect of the pokemon, and not to just reset the map
- change handling of protosynthesis/quark drive too since we store multiple effects due to duplicate showdown protocols. We should remove them all when one gets removed

PR #647, opened 2024-11-21
Adding support for Temporary Types and Abilities

caymansimpson:
Added support for Skill Swap and TypeChange; this allows for us to better track an `AbstractBattle` state that we weren't tracking before. Changes:
1. Added `_temporary_types` internal variable to `Pokemon` that overrides a mon's type (and type-related properties) until switched out. Also changed `_temporary_types` to a `List` because the gen7 exclusive move [Trick-or-Treat](https://bulbapedia.bulbagarden.net/wiki/Trick-or-Treat_(move)) adds typing to a mon's existing typing. Not supported now, but can be extended in the future.
3. Added `_temporary_ability` internal variable to `Pokemon` that overrides a mon's ability until switched out
4. Added parsing of `skillswap` to both update both mon's abilities, but to also update the mon's innate ability if previously unknown
5. Added support for parsing of ability-changing moves like [`worryseed`, `entrainment`, `doodle` and `roleplay`](https://bulbapedia.bulbagarden.net/wiki/Category:Ability-changing_moves) as defined in [showdown protocol](https://github.com/smogon/pokemon-showdown/blob/4cece4ee981d90b26c53ff28a0da9d0def84486b/data/moves.ts#L3956)
6. Added support for parsing of ability-changing moves like [`burnup`, `doubleshock`, ` and `roost`](https://bulbapedia.bulbagarden.net/wiki/Type#Typeless) as defined in [showdown protocol](https://github.com/smogon/pokemon-showdown/blob/4cece4ee981d90b26c53ff28a0da9d0def84486b/data/moves.ts#L2178)
7. Added associate tests both in `Pokemon` and also via `battle.parse_message()`

Mentioned here: https://github.com/hsahovic/poke-env/issues/629

### DIFF
--- a/src/poke_env/environment/abstract_battle.py
+++ b/src/poke_env/environment/abstract_battle.py
@@ -676,10 +676,7 @@ class AbstractBattle(ABC):
             self.get_pokemon(pokemon).end_effect(effect)
         elif event[1] == "-endability":
             pokemon = event[2]
-            if len(event) > 4 and event[4].startswith("[from] move:"):
-                self.get_pokemon(pokemon).set_temporary_ability(None)
-            else:
-                self.get_pokemon(pokemon).ability = None
+            self.get_pokemon(pokemon).set_temporary_ability(None)
         elif event[1] == "-enditem":
             pokemon, item = event[2:4]
             self.get_pokemon(pokemon).end_item(item)
--- a/src/poke_env/environment/abstract_battle.py
+++ b/src/poke_env/environment/abstract_battle.py
@@ -610,7 +610,10 @@ class AbstractBattle(ABC):
             self.get_pokemon(pokemon).boost(stat, -int(amount))
         elif event[1] == "-ability":
             pokemon, ability = event[2:4]
-            self.get_pokemon(pokemon).ability = ability
+            if len(event) > 4 and event[4].startswith("[from] move:"):
+                self.get_pokemon(pokemon).set_temporary_ability(ability)
+            else:
+                self.get_pokemon(pokemon).ability = ability
         elif split_message[1] == "-start":
             pokemon, effect = event[2:4]
             pokemon = self.get_pokemon(pokemon)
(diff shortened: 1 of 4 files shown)


src/poke_env/environment/abstract_battle.py
```diff
@@ -665,7 +676,10 @@ def parse_message(self, split_message: List[str]):
             self.get_pokemon(pokemon).end_effect(effect)
         elif event[1] == "-endability":
             pokemon = event[2]
-            self.get_pokemon(pokemon).ability = None
+            if len(event) > 4 and event[4].startswith("[from] move:"):
+                self.get_pokemon(pokemon).set_temporary_ability(None)
+            else:
+                self.get_pokemon(pokemon).ability = None
```
hsahovic, line 682: What are examples of this path being used?

caymansimpson: For the if: This is tested in test_battle on line 576 (not sure how to link in the comment). `-endability` occurs whenever a `move` removes an ability, according to [showdown protocol](https://github.com/smogon/pokemon-showdown/blob/f755b2852e85715b51895ce92bb165bc58652baa/sim/pokemon.ts#L1838). It triggers first an `-endability` and then an `-ability`, as tested.

For the else: I'm not sure when `-endability` can be used to actually remove a pokemon's permanent ability. Right now, likely Gastro Acid or Neutralizing Gas triggers `-endability` without the `[from] move:` in the `event`, and this continues to be mishandled? But:
1. I'm not clear whether `-endability` should ONLY be used to set `mon._temporary_ability`
2. This is strictly less broken than before :P

If you'd like, I could do some research and circle back? If I do find that there are no permanent ways to end mons' abilities, then I can remove this `if` and create a constant within `Pokemon` (e.g. `NO_ABILITY`) to distinguish between when `Pokemon._temporary_ability` is not set and when a mon should have no ability. I'm just not confident that I can say there exists no edge-cases like this.

hsahovic: Yeah let's double check - if you don't find anything we can merge this PR

caymansimpson: OK, looked through and I don't think there are any ways to permanently remove a mon's ability. So I took the route above and made a constant to track whether a mon is supposed to have no ability (eg removed by `Gastro Acid`) vs when we dont know one. This is abstracted away outside of this class.

Further Fuzz Tested 10K games and it seems like there were no edge-cases that I'm breaking. Also added some testing (and cleaned up syntax in `test_battle` as well) for temporary abilities

hsahovic (approved):

PR #653, opened 2024-12-05
Make some methods of Player static

cameronangliss:


cameronangliss:
@hsahovic I'm seeming to fail CI here but I don't see how my changes would be making this error happen. Is this me?

hsahovic (approved):

PR #654, opened 2024-12-05
Env can accept open sheets

cameronangliss:


hsahovic (approved):
Thanks, this is a great change!

PR #657, opened 2024-12-05
Improved `battle_against` API

cameronangliss:


cameronangliss:
@hsahovic this one's ready for you!

src/poke_env/player/player.py
```diff
@@ -692,7 +694,9 @@ async def _ladder(self, n_games: int):
             perf_counter() - start_time,
         )
 
-    async def battle_against(self, opponent: "Player", n_battles: int = 1):
+    async def battle_against(
```
hsahovic, line 697: Can you update the docstring?

hsahovic: Does this make `n_battles` a required named argument? If so i'd prefer `opponent` to be either a player or a list of players, and keep `n_battles` from requiring to be named.

src/poke_env/player/player.py
```diff
@@ -702,17 +706,30 @@ async def battle_against(self, opponent: "Player", n_battles: int = 1):
...
+                self.send_challenges(
+                    to_id_str(opponent.username),
+                    n_battles,
+                    to_wait=opponent.ps_client.logged_in,
+                ),
+                opponent.accept_challenges(
+                    to_id_str(self.username), n_battles, opponent.next_team
+                ),
+            )
+            results[opponent.username] = (self.win_rate, opponent.win_rate)
```
hsahovic, line 729: This should bot be part of this method - reporting results is unrelated to starting battles. Similarly, resetting the stored battles after the battles are done would be counterintuitive and make for a bad API.

hsahovic (changes requested):

src/poke_env/player/player.py
cameronangliss: Good point! Here you go: f17c415

cameronangliss:
@hsahovic alright I've attended to your thoughts, let me know what you think! Got some nice improvements in, thanks for the review!

hsahovic (approved):

hsahovic:
Thanks @cameronangliss !

PR #658, opened 2024-12-05
Cover neutral gender in parser

cameronangliss:


hsahovic (changes requested):
How did this come up?
Can you add an unit test for this?

cameronangliss:
I'm now thinking this change isn't really in line with how showdown sends messages, so I'm going to close this.

PR #659, opened 2024-12-05
More robust battle tracking

### LINKED ISSUE #616: force_switch is forced before the events of a battle are processed

```in choose_move on turn #1; force_switch: [False, False].

### PR DESCRIPTION
**Pull request by cameronangliss**
Fixes #616

1. Currently we consume the messages in the order they come in, which is request and then protocol message. We really should update the battle object with the protocol first, and then update with the battle request - and now we do! This is better because it prevents us from accidentally overwriting something that is definitely true with our interpretation of the protocol, which currently has bugs. Specifically, this guarantees that we can correctly track Zoroark if it's on our team, and it resolves several bugs in the parsing logic that were not causing errors up to this point.
1. There is no way to send pass in poke-env. Having DefaultBattleOrder as the default option for the orders in DoubleBattleOrder is misleading, and it really should be pass. The user could just provide DefaultBattleOrder to the DoubleBattleOrder constructor if they want one of the sides to be default, so this is a strict empowerment for the user.

### DIFF
--- a/src/poke_env/player/baselines.py
+++ b/src/poke_env/player/baselines.py
@@ -1,5 +1,5 @@
 import random
-from typing import List
+from typing import List, Optional
 
 from poke_env.environment.abstract_battle import AbstractBattle
 from poke_env.environment.double_battle import DoubleBattle
@@ -29,7 +29,7 @@ class MaxBasePowerPlayer(Player):
         return self.choose_random_move(battle)
 
     def choose_doubles_move(self, battle: DoubleBattle):
-        orders: List[BattleOrder] = []
+        orders: List[Optional[BattleOrder]] = []
         switched_in = None
 
         if any(battle.force_switch):
@@ -51,7 +51,7 @@ class MaxBasePowerPlayer(Player):
             switches = [s for s in switches if s != switched_in]
 
             if not mon or mon.fainted:
-                orders.append(DefaultBattleOrder())
+                orders.append(None)
                 continue
             elif not moves and switches:
                 mon_to_switch_in = random.choice(switches)
--- a/src/poke_env/player/battle_order.py
+++ b/src/poke_env/player/battle_order.py
@@ -82,9 +82,9 @@ class DoubleBattleOrder(BattleOrder):
                 + self.second_order.message.replace("/choose ", "")
             )
         elif self.first_order:
-            return self.first_order.message + ", default"
+            return self.first_order.message + ", pass"
         elif self.second_order:
-            return self.second_order.message + ", default"
+            return "/choose pass, " + self.second_order.message.replace("/choose ", "")
         else:
             return self.DEFAULT_ORDER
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Review by hsahovic — CHANGES_REQUESTED** [review]
This is a good change.
Can you add some unit tests?

**Comment by cameronangliss** [comment]
@hsahovic this is being blocked by bugs in poke-env not being able to handle Zoroark properly😞 all other tests pass. I guess the good news is that this PR exposes a bug, but the bad news is that Zoroark is a nightmare. @kman if you're interested in taking down one of the most infamous problems of all time, you're invited hahaha

**Comment by caymansimpson** [comment]
haha I don't think it's worth our time to tackle Illusion; it involves keeping multiple possible battle states and integrating speed/move inferences into the core of poke-env, which if comprehensively done, will require 10K+ lines of code (my inferences work is 95% accurate at 6K lines; there are so many niche interactions).

File: src/poke_env/environment/double_battle.py [hunk 1]
```diff
@@ -130,6 +130,13 @@ def parse_request(self, request: Dict[str, Any]) -> None:
         if side["pokemon"]:
             self._player_role = side["pokemon"][0]["ident"][:2]
         self._update_team_from_request(side)
+        if self.player_role is not None:
```
hsahovic, line 133: why is this change in this PR?

cameronangliss: So, this was a little while ago, but if I remember right, this was to try and simplify the code and fix Zoroark bugs on our side of the battle specifically (since that's all the request tells us about, is our own side).

cameronangliss: to be clear, it's a replacement for the code being removed a few lines further below.

cameronangliss: I definitely could test taking it out real quick and see what happens, whether that breaks stuff.

cameronangliss: Alright, I remember it all now. So, yes, this is a necessary fix, and it is for Zoroark. This is also just a much easier code chunk to look at than what it's replacing. Basically, the request will always put the active pokemon in the first and second positions in the pokemon list, and although you could be worried that we aren't correctly filtering out fainted pokemon, if you look at the _get_active_pokemon method in double_battle.py (which is used to give the active_pokemon property), you'll see that it is filtered there, so we were actually double-filtering before.

hsahovic:
I think this accidentally got commits from https://github.com/hsahovic/poke-env/pull/666?

cameronangliss:
@hsahovic I put #666 into this PR because it ended up that this PR needed the changes from it, and also it needed the changes from this PR in order to handle Zoroark better. I've rebranded this as an effort to improve the battle tracking in poke-env. Let me know what you think!

src/poke_env/environment/double_battle.py
cameronangliss: I've now moved this change to #666, as that is where it is more relevant.

PR #664, opened 2024-12-22
PettingZoo Integration

### LINKED ISSUE #560: [Proposal] Replace `gymnasium` with `pettingzoo`
Pokemon battles are fundamentally a multi-agent system.

Currently we are using [Farama's](https://farama.org/) [Gymnasium](https://gymnasium.farama.org/) which is a RL env API for _single-agents_. I believe this decision came about as at the time of this project's creation there was no standard/popular multi-agent RL env API, leaving us to choose [OpenAI's](https://openai.com/) [Gym](https://github.com/openai/gym), which has now moved over to Gymnasium.

However, with popularity rising in multi-agent systems, various RL env API's have been created for multi-agent systems, such as Farama's [PettingZoo](https://pettingzoo.farama.org/) and Google Deepmind's [OpenSpiel](https://github.com/google-deepmind/open_spiel) among others.

Overall, PettingZoo has become a strong defacto for _multi-agent_ systems as it closely resembles Gymnasium's API, is mathematically more general than the other RL multi-agent env APIs, and more importantly has seen adoption in the popular RL libraries such as [rllib](https://pettingzoo.farama.org/tutorials/rllib/), [tianshou](https://pettingzoo.farama.org/tutorials/tianshou/), [stable-baselines3](https://pettingzoo.farama.org/tutorials/sb3/), among others.

With this in mind, and again recognizing that Pokemon battles are a _multi-agent_ system, I believe that we should be using `pettingzoo` instead of `gymnasium`.

Practically, this would mean a change in the API for us and it should mean that certain logic we may have assumed/worked around due to using a single-agent API would be simpler in PettingZoo's API.

I'd love to hear everyone's thoughts about this and the feasibility of this change. I truly believe such a change would increase the usability of this project.

cameronangliss:
Fixes #560

This is the minimal code-change I could manage in order to integrate PettingZoo with poke-env. That being said, there are a lot of optimizations and improvements to the env code that I wanted to put in here, but decided that it would be better to leave this as minimal as possible, as the diff is substantial already. I've started working on such improvements, which can be tracked [here](https://github.com/cameronangliss/poke-env/compare/pettingzoo...cameronangliss:poke-env:optimize-env?expand=1).


src/poke_env/player/gymnasium_api.py
```diff
@@ -270,149 +305,161 @@ def action_space_size(self) -> int:
...
-        battle.logger = None
-        self.last_battle = battle
-        self._actions.put(action)
-        observation = self._observations.get()
-        reward = self.calc_reward(self.last_battle, self.current_battle)
+        battle1 = copy.copy(self.current_battle1)
+        battle1.logger = None
+        battle2 = copy.copy(self.current_battle2)
+        battle2.logger = None
+        self.last_battle1 = battle1
```
caymansimpson, line 371: Sorry for the random (and potentially ignorant) question - do we need to store `last_battle`s in the env? I ask for two reasons:
1. My understanding is `copy.copy` won't deep copy `Pokemon` objects, so `calc_reward` won't work properly because the Pokemon objects will be the same across battle copies (e.g. if a pokemon faints in this turn recorded by `battle1`, it will also appear fainted in `last_battle1`); I don't think this is tested either. Deep Copy is also suuuuuper slow, and so there could be a big tradeoff here
2. IIUC, the difference in reward isn't a necessary reward design; e.g. it makes sense to just reward agents for reaching an end state. In this case, it would make sense to allow the user to save `last_battle` in their own custom `Player` class and access it when they need to define the reward as a difference between two states

Im also not very confident in either #1 or #2, so sorry if this is random noise haha

cameronangliss: I actually agree entirely. My reasoning is that if someone wants the last battle, they can store it themselves, we shouldn't be going through the trouble. That's coming in the follow-up PR :) I just wanted to make this diff as minimal and trustable as possible.

hsahovic (approved):

hsahovic:
Can you resolve the conflict? Other than that this is great work - I'm excited to try it out!

hsahovic:
Great job @cameronangliss - thanks!!!

PR #667, opened 2024-12-28
Revival Blessing Bugfix

cameronangliss:
Fixes bug where revival blessing heals pokemon, but its faint status is not cleared. We fix this by making it so that any time `.heal()` is called on a fainted pokemon object, it always clears the `_status` field.


hsahovic:
Thanks @cameronangliss !

PR #668, opened 2024-12-30
PettingZoo II + env refactor and cleanup

cameronangliss:
This gets the entire RL environment situation going on in poke-env to a point that I feel is much better. This was a lot! But a lot of this is interdependent and I'm not sure how much I can split it up into differrent PR's, so let's just see how this goes and sorry for the giant load!

Changes include (broadly):
1. Introducing `SingleAgentWrapper`, a wrapper to be used over environments that allows users of poke-env to still have the way things used to work before pettingzoo integration, where you specify the opponent as a player object and only send battle commands in for one side of the battle when interacting via the environment.
1. `EnvPlayer` is gone, as well as all `Gen{i}EnvSinglePlayer` classes, in favor of `SinglesEnv` and `DoublesEnv` (yes, we support doubles now). All methods that were unique to `EnvPlayer`, as well as the unique things that happened in its `__init__` method, were migrated to `GymnasiumEnv` (which is now renamed to `PokeEnv`).
1. `action_to_order` is now defined in `SinglesEnv` and `DoublesEnv` in a way that is generalized to any generation of Pokemon (as long as it's >= gen4). Also, an inverse `order_to_action` method has been added, which should be a useful utility for the user, and was necessary for me to create the aforementioned `SingleAgentWrapper`.


src/poke_env/player/doubles_env.py
```diff
@@ -0,0 +1,273 @@
...
+        except IndexError:
+            return Player.choose_random_move(battle)
+        except AssertionError as e:
+            if str(e) == "invalid pick":
+                return Player.choose_random_move(battle)
+            else:
+                raise e
+
+    @staticmethod
+    def _action_to_order_individual(
```
caymansimpson, line 154: Would it be better to have a map that has bidirectional mapping between order and int? That way we don't have to store gimmicks, and it reduces complexity and difficulty to parse this

src/poke_env/player/doubles_env.py
```diff
@@ -0,0 +1,273 @@
...
+            accept_open_team_sheet=accept_open_team_sheet,
+            start_timer_on_battle_start=start_timer_on_battle_start,
+            start_listening=start_listening,
+            open_timeout=open_timeout,
+            ping_interval=ping_interval,
+            ping_timeout=ping_timeout,
+            team=team,
+            start_challenging=start_challenging,
+        )
+        num_switches = 6
```
caymansimpson, line 61: Should we leave this as a param? For VGC this is not the case; this is 4. Or at least, extend the parsing of battle_format so it takes into account vgc

src/poke_env/player/doubles_env.py
```diff
@@ -0,0 +1,273 @@
...
+            num_gimmicks = 4
+        else:
+            num_gimmicks = 0
+        act_size = num_switches + num_moves * num_targets * (num_gimmicks + 1)
+        self.action_spaces = {
+            agent: MultiDiscrete([act_size, act_size]) for agent in self.possible_agents
+        }
+
+    @staticmethod
+    def action_to_order(
```
caymansimpson, line 80: For this, I see you call have a bunch of asserts. Should they be handled by action masking by the end user? There are a bunch of other cases where there are illegal moves too that aren't covered here, so it's a bit weird to only cover some cases in these internal methods

cameronangliss: Yes! I completely agree. I left it like this because the way poke-env worked before was like this, and so I just wanted to not change that to make a smooth transition. You're right though, at some point a PR should be made where the player doesn't have to rely on the "do a random move if my move doesn't make sense" flow.

src/poke_env/player/doubles_env.py
cameronangliss: Do you mean just a bidirectional mapping for gimmick specifically, or from any order to any action? Now that you mention it, it's probably possible to make a function that is battle object -> a mapping of action indices to orders. That would be an extremely useful utility function, and could be definitely used to significantly reduce complexity of this code.

src/poke_env/player/doubles_env.py
cameronangliss: @caymansimpson could you tell me what edge cases I'm not asserting away? In the integration tests I am not seeing that any invalid moves are made, so I think everything is covered as far as I can tell

caymansimpson: in _action_to_order_individual, I think you're missing whether we're trying to switch to the same mon, whether we're trying to do two dynamaxes/megas/teras/z-moves and whether we're sending an order for a mon that is fainted/None

I have an attempt at a full list here: https://github.com/caymansimpson/EliteFurretAI/blob/development/src/elitefurretai/utils/battle_order_validator.py

cameronangliss: No, we have those. Just look at each assert in the _action_to_order_individual method. For example, we know the switch is legal because I assert the following: `assert order.order in battle.available_switches[pos], "invalid pick"`.

caymansimpson: Ah maybe it will be helpful to add unit_tests? Or could you point to the line that is supposed to cover these?

From what I understand, if you are in a gamestate where:
1. You have two mons out in field (1, 2 indicies in showdown protocol)
2. Two mons in back that are available to switch (3, 4 in showdown protocol)

Available switches in the request will point to the mons: [[3, 4], [3, 4]]. So this check isn't sufficient because according to mon.available_switches, you can trying to switch to the 3rd index twice would still pass (e.g. /choose switch 3, switch 3)

Similarly, if you are in a state where both your mon can tera, you will have battle.can_tera be [true, true] so in this code you can return ("/choose move 1 tera, move 1 tera") which is invalid? (plus the other things I called out above)

Agree that these should be caught by integration tests; Im looking at the code for all the integration tests, and they're all for singles, and there are no doubles integration tests. I think we're introducing bugs here and so maybe we should also add integration tests for doubles?

cameronangliss: Your concern about both pokemon switching to the same pokemon in available_switches is covered by the use of join_orders. this also covers two teras, two megas, etc. Basically ensures that both orders can actually be made simultaneously.

caymansimpson: Ahhhh, missed those! Ignore me!

src/poke_env/player/doubles_env.py
cameronangliss: @caymansimpson I looked into this and decided probably it would be best to leave as a follow-on PR or something. I made #683 which breaks off a big chunk of this PR (just the SinglesEnv stuff) so we can keep talking there. There are honestly several improvements to be made off of where this is (such as adding an action mask so that we don't need all of these asserts for example).

src/poke_env/player/doubles_env.py
cameronangliss: I'm open to this. Let's discuss it after #683 merges in and I make a break-off PR specifically for the DoublesEnv.

cameronangliss:
I made this a draft since I'm breaking this up into multiple smaller PRs at this point, but leaving it for now just for reference and comment history.

PR #670, opened 2025-01-03
Add focus punch to _FROM_DATA

cameronangliss:
We are missing focus punch in the _FROM_DATA list.


hsahovic (approved):

hsahovic:
good catch @cameronangliss - thanks!

PR #671, opened 2025-01-14
Get rid of last_battle

cameronangliss:
This should help some with the diff of #668.

I removed `last_battle` as a tracked entity in the env. This is because if the user of poke-env wants to track the last battle object, they can copy it for themselves, but we shouldn't be doing it if the user doesn't want that. This allowed me to rename `current_battle` -> `battle` all over the code in gymnasium_api.py.


cameronangliss:
@hsahovic this feels like an easy merge to me, but lmk what you think :)

hsahovic (approved):

PR #674, opened 2025-01-16
No more `describe_embedding`

cameronangliss:
This should help ease down the diff of #668.

`describe_embedding` is no longer the way that the user informs the env class about the embedding space. The user will now do so inside of the `__init__` method by assigning the `observation_spaces` attribute.


hsahovic (commented):
I think this goes in the right direction - having to override a method might be too much.

However, i'm not sure making the recommended way of specifying `describe_embedding` an `__init__` override is better - do you think a simpler approach of just setting the attribute could work?

That wouldn't preclude going the `__init__` override route but might be simpler for users.

cameronangliss:
@hsahovic the user can set the attribute outside of the `__init__` if they want. The only reason I'm doing it after doing `super().__init__()` is because the way I'm making the observation_spaces, I need the possible_agents field that is defined in the superclass. Theoretically, the user doesn't even technically need to define observation_spaces to use the environment, but it just won't pass `parallel_api_test`.

PR #683, opened 2025-01-28
Introduce `SinglesEnv`

cameronangliss:
Requires #674, #678, #695

1. `EnvPlayer` is gone, as well as all `Gen{i}EnvSinglePlayer` classes, in favor of `SinglesEnv`. All methods that were unique to `EnvPlayer`, as well as the unique things that happened in its `__init__` method, were migrated to `GymnasiumEnv` (which is now renamed to `PokeEnv`, and gymnasium_api.py is now named env.py).
1. All `action_to_order` methods from the `Gen{i}EnvSinglePlayer` classes have been condensed into a single method that is defined in`SinglesEnv`, and an inverse method called `order_to_action` was also included.
1. Because `SinglesEnv` has a parameterized generation number, I was able to vastly simplify the code in integration_tests/test_env_player.py (renamed to test_env.py) and unit_tests/player/test_gymnasium.py and unit_tests/player/test_env_player.py (condensed into a single test_env.py file).


docs/source/modules/player.rst
hsahovic, line 1: Do you want to add docs for the new classes?

integration_tests/test_env.py
```diff
@@ -0,0 +1,80 @@
...
+        env = SinglesTestEnv(
+            battle_format=f"gen{gen}randombattle",
+            log_level=25,
+            start_challenging=False,
+            strict=False,
+        )
+        env.start_challenging(3)
+        play_function(env, 3)
+        env.close()
+    env = SinglesTestEnv(
```
hsahovic, line 46: Can you separate the repeated call tests in a separate function?

examples/env_example.py
hsahovic, line 1: Maybe for a separate PR, but maybe a more realistic example would be useful as a notebook :)

src/poke_env/player/singles_env.py
```diff
@@ -0,0 +1,233 @@
...
+from poke_env.player.player import Player
+from poke_env.ps_client import (
+    AccountConfiguration,
+    LocalhostServerConfiguration,
+    ServerConfiguration,
+)
+from poke_env.teambuilder import Teambuilder
+
+
+class SinglesEnv(PokeEnv[ObsType, np.int64]):
```
hsahovic, line 22: What do you think of adding a default `calc_reward`?

unit_tests/player/test_env.py
```diff
@@ -1,58 +1,125 @@
...
-
-    def action_to_order(self, action: np.int64, battle: AbstractBattle) -> BattleOrder:
-        return Gen7EnvSinglePlayer.action_to_order(self, action, battle)
-
-    _ACTION_SPACE = Gen7EnvSinglePlayer._ACTION_SPACE
-
-    def embed_battle(self, battle):
-        return None
+class CustomEnv(SinglesEnv[npt.NDArray[np.float32]]):
+    def __init__(self, *args, **kwargs):
```
hsahovic, line 28: is this `__init__` required?

hsahovic (changes requested):

integration_tests/test_env.py
cameronangliss: cb335bf6f67961f59ed9cafcb0860c94a3152f69

unit_tests/player/test_env.py
cameronangliss: e10ac6f49f4b70bc846eaa84611d200bad80747e

docs/source/modules/player.rst
cameronangliss: 5f4f17940a491396095dcbf0b169955420ce7024

examples/env_example.py
cameronangliss: Agreed! I just basically copied the way that we were already doing things in gymnasium_example.py

src/poke_env/player/singles_env.py
cameronangliss: What do you think of just using reward_computing_helper by default: 1fc6b291aad186c6e5f75fb912420ab841320e3c

cameronangliss: I actually removed this change, I don't think I like it so much. I think it's a good thing to force the user to specify embed_battle and calc_reward explicitly, and we have reward_computing_helper in `PokeEnv` for them to use if they wish (which is also shown in the example in env_example.py).

hsahovic (approved):

PR #686, opened 2025-02-02
Introduce `SingleAgentWrapper`

cameronangliss:
Requires #704

Introducing `SingleAgentWrapper`, a wrapper to be used over environments that allows users of poke-env to still have the way things used to work before pettingzoo integration, where you specify the opponent as a player object and only send battle commands in for one side of the battle when interacting via the environment.


integration_tests/test_env.py
```diff
@@ -82,3 +97,7 @@ def test_env_api():
         )
         parallel_api_test(env)
         env.close()
+        env = SingleAgentWrapper(env, RandomPlayer())
```
hsahovic, line 100: Maybe this should be in a separate test?

integration_tests/test_env.py
```diff
@@ -40,8 +51,12 @@ def test_env_run():
             start_challenging=False,
             strict=False,
         )
-        env.start_challenging(3)
-        play_function(env, 3)
+        env.start_challenging(10)
+        play_function(env, 10)
+        env.close()
+        env = SingleAgentWrapper(env, RandomPlayer())
```
hsahovic, line 57: Maybe this should be in a separate test?

hsahovic (approved):

integration_tests/test_env.py
cameronangliss: f561085acec931fa6548af0f550383fb385a3114

integration_tests/test_env.py
cameronangliss: f561085acec931fa6548af0f550383fb385a3114

PR #687, opened 2025-02-02
Introduce `DoublesEnv`

cameronangliss:
This introduces a sister class to `SinglesEnv`, providing a way to train an RL agent in the doubles format!


caymansimpson:
Amazing - ill take a look hopefully tonight

integration_tests/test_env.py
```diff
@@ -4,7 +4,7 @@
 from gymnasium.utils.env_checker import check_env
```
hsahovic: the logic looks good, but i think it would make more sense to split this file into `split_singles_env` and `test_doubles_env`

hsahovic (approved):

integration_tests/test_env.py
cameronangliss: did it!

PR #695, opened 2025-02-07
Fix concurrency bug

cameronangliss:
This fixes a bug caused by an occasional misordering of player role message when using concurrency. The battle object can now infer its player role, whether it receives information about its own player role or the opponent's player role first. I also made a small refactor in the get_pokemon method to make the logic more explicit and readable.


cameronangliss:
@hsahovic I believe this bug is fixed now with these changes, this one is ready to go.

hsahovic (approved):
Great catch!

PR #704, opened 2025-02-23
Fix several bugs getting caught by CI

cameronangliss:
1. Player now tracks a `_waiting` Event object as well as the `_trying_again` Event object. If the `_EnvPlayer` instance has `_waiting` set or if the opposing _EnvPlayer instance has `_trying_again` set, then it will neglect to follow the order passed in through the `action` parameter of `step()`. This `Event`-based handling of choosing whether or not to listen to the action passed to `step()` resolves exceedingly rare race conditions (so rare that I never once saw it show up in CI, I saw this when using poke-env for my own personal project).
1. Player tracks `_trying_again` more correctly, and add code to make maybe_default_order lead to 100% default chance if we have a pokemon on our team with illusion which could be messing up our knowledge about what actions are legal.
1. Forfeiting logic is fortified to guarantee that it works even if one of the players is trying again after an erroneous action attempt or if one of the players is waiting.
1. Renamed `get_timeout` to `race_get`, refactored it, and generalized it to be able to take in an arbitrary number of Event objects.
1. There was an issue with how we were parsing messages which was causing an issue in `abstract_battle.py`, and I fixed that as well.
1. Added a much-needed unit test for reset(), step(), and close() in PokeEnv.
1. A few simplifications and renamings to make the code logic clearer.

@hsahovic let me know what you think!

### DIFF
--- a/src/poke_env/player/env.py
+++ b/src/poke_env/player/env.py
@@ -43,17 +43,17 @@ class _AsyncQueue(Generic[ItemType]):
         return res.result()
 
     def get_unless_waiting(self, battle: AbstractBattle) -> Optional[ItemType]:
-        print(battle.player_username, "searching")
+        print(battle.player_username, "searching", flush=True)
         while not battle._wait:
             try:
                 res = asyncio.run_coroutine_threadsafe(
                     asyncio.wait_for(self.async_get(), timeout=0.01), POKE_LOOP
                 )
-                print(battle.player_username, "got it")
+                print(battle.player_username, "got it", flush=True)
                 return res.result()
             except asyncio.TimeoutError:
                 continue
-        print(battle.player_username, "gave up")
+        print(battle.player_username, "gave up", flush=True)
         return None
 
     async def async_put(self, item: ItemType):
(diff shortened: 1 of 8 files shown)


src/poke_env/player/env.py
```diff
@@ -326,9 +326,16 @@ def reset(
                 count -= 1
                 time.sleep(self._TIME_BETWEEN_RETRIES)
         if self.battle1 and not self.battle1.finished:
+            assert self.battle2 is not None
             if self.battle1 == self.agent1.battle:
-                self.agent1.order_queue.put(ForfeitBattleOrder())
-                self.agent2.order_queue.put(DefaultBattleOrder())
+                if self.agent1.waiting and not (self.battle1._wait or False):
```
hsahovic, line 331: why is the `or False` required?

cameronangliss: Good point, I was just doing that to make it a boolean in case it's None, but the logic would be the same without. I removed it.

cameronangliss:
I'm giving this my official sign-off! I ran it overnight with 4 parallel RL runs and had no issues.

cameronangliss:
I just added a bunch of new commits just because I was trying to raise the test coverage since it was getting a little low after this PR's changes. This is still ready to go though.

hsahovic (approved):

PR #709, opened 2025-03-04
Replace `Snow` with `Snowscape`

shrianshChari:
Resolves compatibility issues created by https://github.com/smogon/pokemon-showdown/commit/deda1db857f5261c470bd262da7a5fd69f3b78e8

Alternatively, we can edit the `fromshowdownmessage` method to try and turn messages that are read in as "Snowscape" to also return `Weather.SNOW`, which would maintain compatibility with older builds of the Showdown server.


hsahovic:
Thanks @shrianshChari ! I agree we could try keeping this backwards-compatible. I will merge this to fix the current situation, but If you want to follow up with the other change, I'll happily merge it as well.

Issue #708, opened 2025-02-28
Error messages after using the moves Reflect Type or Heart Swap


LordZapan:
I'm getting the following error messages every time a RandomPlayer uses the Reflect Type or Heart Swap (or when Metronome lands on one of those two moves) and the blattle will stall at the beginning of the turn.

I'm doing RandomPlayer vs RandomPlayer in Gen 9 custom games.
Any idea what is going on?

```
D:\Download\poke-env\PokeFormated.py:44: DeprecationWarning: There is no current event loop
  asyncio.get_event_loop().run_until_complete(main())
2025-02-27 16:18:15,040 - RandomPlayer 1 - ERROR - Unhandled exception raised while handling message:
>battle-gen9customgame-7670
|
|t:|1740669495
|move|p1a: Tentacruel|Reflect Type|p2a: Ponyta
|-start|p1a: Tentacruel|typechange|[from] move: Reflect Type|[of] p2a: Ponyta
|move|p2a: Ponyta|Tackle|p1a: Tentacruel
|-damage|p1a: Tentacruel|311/348
|
|upkeep
|-start|p1a: Tentacruel|typechange|Fire|[silent]
|turn|2
Traceback (most recent call last):
  File "C:\Users\lordz\AppData\Local\Programs\Python\Python310\lib\site-packages\poke_env\ps_client\ps_client.py", line 136, in _handle_message
    await self._handle_battle_message(split_messages)  # type: ignore
  File "C:\Users\lordz\AppData\Local\Programs\Python\Python310\lib\site-packages\poke_env\player\player.py", line 370, in _handle_battle_message
    battle.parse_message(split_message)
  File "C:\Users\lordz\AppData\Local\Programs\Python\Python310\lib\site-packages\poke_env\environment\abstract_battle.py", line 622, in parse_message
    pokemon.start_effect(effect, details=event[4])  # type: ignore
  File "C:\Users\lordz\AppData\Local\Programs\Python\Python310\lib\site-packages\poke_env\environment\pokemon.py", line 406, in start_effect
    self._temporary_types.append(PokemonType.from_name(type_))
  File "C:\Users\lordz\AppData\Local\Programs\Python\Python310\lib\site-packages\poke_env\environment\pokemon_type.py", line 83, in from_name
    return PokemonType[name.upper()]
  File "C:\Users\lordz\AppData\Local\Programs\Python\Python310\lib\enum.py", line 440, in __getitem__
    return cls._member_map_[name]
[...]

PR #711, opened 2025-03-08
Fix Reflect Type/Heart Swap errors

TVFLabs:
I was playing around with this library (what do they call them in Python? "packages"? "modules"?) and I discovered an issue with the moves Reflect Type and Heart Swap. Judging by issue #708, it seems I'm not the first person to encounter this. I wanted to fix it myself so I could continue with the project I was using poke-env for, and once I had I figured I may as well submit a PR.
Not sure if this is the ideal way to do it, but it does fix the issue, and appears to do so without presenting any new problems.


hsahovic (approved):

hsahovic:
Thanks!

PR #712, opened 2025-03-10
Add backwards compatibility with servers that use Snow as the weather

shrianshChari:


cameronangliss:
Probably we'd want to put the logger part after in an else after that if you added?

hsahovic (approved):

PR #720, opened 2025-04-16
Some more CI failure fixes

cameronangliss:
Hoping this is the last of this type of PR for a little while, and apologies for any issues I've caused!

1. We weren't properly handling the case where both players try again simultaneously after an incorrect action selection in the step() method. This is now fixed. I've also improved the code that handles the race conditions we have dealt with in the past. The new code has the ad hoc property that the order_queue is written to if and only if the battle_queue was just read from, which is guaranteed via the new `agentx_to_move` properties. Readability is also enhanced as a result.
1. I made the code a little more lenient so that it doesn't error out when dealing with frisk. This was causing issues in CI, but I admittedly have seen this cause an issue only once or twice ever (and anyone who knows me knows that I use the CI to validate my code A LOT). Regardless, this is a bit safer for the user.
1. Ramp up the number of battles to run in the integration testing so that issues surface less sporadically.
1. Add some more assert statements to the conversion functions to make sure action_to_order and order_to_action mirror each other more exactly and also catch "trapped" instances when trying to switch.

### DIFF
--- a/integration_tests/test_env.py
+++ b/integration_tests/test_env.py
@@ -31,7 +31,7 @@ def play_function(env, n_battles):
             done = any(terminated.values()) or any(truncated.values())
 
 
-@pytest.mark.timeout(300)
+@pytest.mark.timeout(600)
 def test_env_run():
     for gen in range(4, 10):
         env = SinglesTestEnv(
@@ -40,8 +40,8 @@ def test_env_run():
             start_challenging=False,
             strict=False,
         )
-        env.start_challenging(100)
-        play_function(env, 100)
+        env.start_challenging(200)
+        play_function(env, 200)
         env.close()
 
 
--- a/integration_tests/test_env.py
+++ b/integration_tests/test_env.py
@@ -31,7 +31,7 @@ def play_function(env, n_battles):
             done = any(terminated.values()) or any(truncated.values())
 
 
-@pytest.mark.timeout(120)
+@pytest.mark.timeout(300)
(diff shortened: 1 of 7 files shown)


src/poke_env/player/singles_env.py
```diff
@@ -112,6 +112,7 @@ def action_to_order(
             elif action < 6:
                 order = Player.create_order(list(battle.team.values())[action])
                 if not fake:
+                    assert not battle.trapped, "invalid action"
```
hsahovic, line 115: Can you add docstrings for fake and strict here?

integration_tests/test_env.py
```diff
@@ -40,8 +40,8 @@ def test_env_run():
             start_challenging=False,
             strict=False,
         )
-        env.start_challenging(3)
-        play_function(env, 3)
+        env.start_challenging(10)
```
hsahovic, line 43: How long does this add to the CI?

cameronangliss: I haven't measured exactly, but the difference is minimal. It seems most of the time is taken spinning the environment up initially, but actually running battles is fast.

src/poke_env/player/singles_env.py
cameronangliss: sure thing! 1b57b6566bebc38ba3eb9727c2af9cbb3fd6f1bd

hsahovic (approved):

PR #729, opened 2025-05-03
Tweak for running the basic example and logging potential improvement

Gummygamer:
I had trouble running the examples with the current version of the codebase, please check whether the commits contain relevant changes. The logging should be useful, at least if it is triggered in debug mode. The change of making the strict argument false in env.py can be the most problematic, but I needed it to have the basic example at least having a forfeit and battles finishing.

### DIFF
--- a/examples/env_example.py
+++ b/examples/env_example.py
@@ -2,9 +2,14 @@ import numpy as np
 import numpy.typing as npt
 from gymnasium.spaces import Box
 from pettingzoo.test.parallel_test import parallel_api_test
+from typing import Any, Dict, Tuple, TypeVar
 
 from poke_env.environment.abstract_battle import AbstractBattle
 from poke_env.player import SinglesEnv
+
+ObsType = TypeVar("ObsType")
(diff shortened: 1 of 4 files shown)


cameronangliss:
Examples may be out of date, we've been updating a bunch of stuff with the RL environment under the hood with pettingzoo. I'm planning to update the examples once a few more PRs come in so that we can demonstrate how to use the environment without needing to use the `SingleAgentWrapper`.

Gummygamer:
Thanks for the feedback, I can move the change in the strict argument to the example then. I also noted other signs of deprecated code usage in the examples, I might continue checking those.

cameronangliss:
I think #733 could possibly cover the concerns of this PR @hsahovic

hsahovic:
@Gummygamer is this addressed by https://github.com/hsahovic/poke-env/pull/740?
If not, can you rebase this branch and i'll review?

Gummygamer:
I believe @cameronangliss addressed the relevant contributions. I might close the PR for now.

PR #730, opened 2025-05-04
CI Maintenance

### LINKED ISSUE #616: force_switch is forced before the events of a battle are processed

cameronangliss:
Resolves #616
1. gen9vgc2025regg has been deprecated on showdown, so I updated all references to it to "gen9vgc2025regi"
1. Fixed the return type of a property in move.py
1. Showdown is now handling the request/protocol misordering for us (from https://github.com/smogon/pokemon-showdown/commit/091241f1bbb2009b79f8e1a8e24f3b32d4e6150e, yay!) so I updated the code to expect it in the protocol-request order now, rather than the previous order which was request-protocol


hsahovic (approved):

hsahovic:
thank you!!

PR #733, opened 2025-05-25
Add `valid_orders` properties

cameronangliss:
1. Add `valid_orders` methods to Battle and DoubleBattle which can help replace all of the validation logic spread throughout the action_to_order and order_to_action methods of SinglesEnv and DoublesEnv.
1. Made the error messages in the conversion methods be more verbose and useful (inspired by #729).
1. When strict=False for the conversion functions, the default fallback is now to make a random move rather than to DefaultBattleOrder, removing unnecessary bias.
1. The Player object has its random move methods refactored to just call the valid_orders methods.
1. Integration tests now test strict=True and scrict=False. This revealed a bug in the code that determines what available_switches will be where it wasn't handling reviving correctly, and this has been fixed.

### DIFF
--- a/src/poke_env/battle/battle.py
+++ b/src/poke_env/battle/battle.py
@@ -260,7 +260,8 @@ class Battle(AbstractBattle):
         orders: List[SingleBattleOrder] = []
         if self._wait:
             return [DefaultBattleOrder()]
-        orders += [SingleBattleOrder(mon) for mon in self.available_switches]
+        if not self.trapped:
+            orders += [SingleBattleOrder(mon) for mon in self.available_switches]
         if self.active_pokemon is None or self.force_switch:
             return orders
         orders += [SingleBattleOrder(move) for move in self.available_moves]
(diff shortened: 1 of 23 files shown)


src/poke_env/battle/battle.py
```diff
@@ -261,3 +254,33 @@ def trapped(self) -> bool:
     @trapped.setter
     def trapped(self, value: bool):
         self._trapped = value
+
+    @property
+    def valid_orders(self) -> List[SingleBattleOrder]:
+        orders: List[SingleBattleOrder] = []
+        if self._wait:
+            return [DefaultBattleOrder()]
+        orders += [SingleBattleOrder(mon) for mon in self.available_switches]
```
hsahovic, line 263: Should this check for `trapped`?

hsahovic (approved):

src/poke_env/battle/battle.py
cameronangliss: I'll look into that. I tried looking into that before and found that it was fine without but I would prefer it be in there as well. I'll put this into draft until I figure that out.

cameronangliss: f8af24cb62be479b5273b919ea8e85c9a3fa861b

hsahovic (approved):

PR #763, opened 2025-08-05
Dondonzo, SymbiosisFairy Lock, Plus/Minus and BattleOrder __eq__ support. Changed Teampreview Storage

caymansimpson:
This PR has quite a bit:
1. Added support for Dondozo given RegH is coming. Successfully removes COMMANDER effects from Tatsugiri. I don't remove Tatsugiri from possible_showdown_targets because it's still targetable in-game even if you can't hit it.
2. Added FairyLock support, which is not supported in Gen9, but isn't currently implemented
3. Added handling for Symbiosis, which now throws an error and doesn't correctly handle item transferrance
4. Corrected an edge case in damage calc for Plus/Minus
5. Simplified some code in damage calc that is unnecessary because we already pass the defender_identifier and attacker_identifier directly
6. Added support for equality checking for BattleOrders. I don't check message so we don't have to construct the string, but can make it simpler if you'd like and implement equality checking using the `message` property
7. Store Teampreview in a List instead of a Set. This is really important because the teampreview order can be sent as a list of numbers, corresponding to the order in which they were sent by showdown logs. This means that order is important to be preserved

Let me know if you have any questions or requested changes! They're all pretty simple


cameronangliss:
Why not just implement `__eq__` by doing `return self.message == other.message`? Also other should be of type BattleOrder.

caymansimpson:
Per my and cameron's comment, just made the __eq__ method more simple.

cameronangliss:
I believe it's technically a guarantee that if the messages are equal then the types must be equal. And I meant you should do other: BattleOrder in the type signature. Not trying to be annoying though haha these are minor details.

hsahovic (approved):

hsahovic:
Thank you!

PR #769, opened 2025-08-21
Added items and ability as gen data + updated data

AlexAuragan:
I wanted to have at least the list of items and abilities for vectorization of observable data for machine learning.
I added scripts/update_items.py and scripts/update_abilities.py with the same format as the other update script.

I think the helper function data_script_utils.py > fetch_and_clean_ps_data() was not working properly on these data but also in moves.py, with the help of LLM I could tweak the regex a bit to have it work on everything currently online. 

I also ran update_all_data.py to update all the data

FYI I'm not yet 100% familiar with everything in the poke-env and showdown ecosystem. In the item data many items have the inherit attribute, Idk if it's related their effect or if it's just a coding artifact that we should get rid of since it doesn't concern the ed user.


src/poke_env/data/static/moves/gen1moves.json
```diff
@@ -573,15 +579,15 @@
...
+        "priority": -5,
         "secondary": null,
         "target": "scripted",
         "type": "Fighting",
         "willCrit": false
     },
     "crabhammer": {
         "accuracy": 85,
-        "basePower": 90,
+        "basePower": 100,
```
hsahovic, line 590: How did that happen? I think this is incorrect - crabhammer had a base power of 90 in gen1

hsahovic (changes requested):
Thanks for opening this PR - this is a good idea.
I think there's something weird about the data cleaning, as it looks like some of the moves were updated with data from later gen (eg. crabhammer with 100 bp in the gen 1 file). Can you fix it?

hsahovic:
I'm going to close this for now - feel free to revive it if you want to address the comments I made above

PR #772, opened 2025-09-01
Gen12damage calc

jkim2729:
I added a damage calculator for gen 1 and 2 along with unit tests. Code is based off the Smogon Damage Calculator Link: https://github.com/smogon/damage-calc/blob/master/calc/src/mechanics/gen12.ts. It returns the min, max, and full list of damage rolls. 

Note that this is not 100% accurate as gen 1 bugs make the pokemon's actual stats not always match what is recorded in poke-env. Pursuit is also not implemented as I do not know how to track swaps.


hsahovic:
Thanks for this change - this is great! Can you make sure the CI passes?

jkim2729:
It should pass tests now.

hsahovic:
Thanks @jkim2729 !

PR #780, opened 2025-09-24
Add VGC teampreview functionality into environments

cameronangliss:


cameronangliss:
@hsahovic this is ready for review whenever you get a chance!

hsahovic:
I could be convinced this makes sense, but my understanding is that this is trying to rely on `choose_move` for `teampreview`, which I'm not convinced is correct. Why do you think we should generally do it that way?

cameronangliss:
@hsahovic I'm doing that because choose_move is gathering decisions from the `_EnvPlayer`s, which is the action input from `env.step()`. Basically this allows the RL agent to control the battle during teampreview using `switch` actions. One could do it differently I suppose but I see no reason why not to do it like this, since the `switch` action translates 1-to-1 to the teampreview decision case (you just need to make two double-switch decisions in a row to get a teampreview decision of length 4).

PR #786, opened 2025-10-19
Fix abilities that copy moves

### LINKED ISSUE #785: Gardevoir tracing Dancer
When Gardevoir has the Trace Ability and the opponent's Oricorio with the ability Dancer uses a move that is affected by Dancer, an `AssertionError` occurs:

`AssertionError: Error with move psyshock. Expected self.moves to contain copycat, metronome, mefirst, mirrormove, assist, transform or mimic. Got {'quiverdance': quiverdance (Move object), 'revelationdance': revelationdance (Move object), 'moonblast': moonblast (Move object), 'mysticalfire': mysticalfire (Move object)}
`

This happens because there are no fallbacks for cases where the Pokémon has moved with its ability. This could be fixed by checking if the ability is Trace or Dancer (maybe Magic Bounce as well).

SantiagoRR2004:
Fixes #785; This is meant to solve the issue I created at #785. I just check if the Pokémon has the ability Dancer or Trace. I don't know if this will solve the problem completely, but it should no longer raise an error.

### DIFF
--- a/src/poke_env/battle/pokemon.py
+++ b/src/poke_env/battle/pokemon.py
@@ -661,7 +661,7 @@ class Pokemon:
                     "transform",
                     "mimic",
                 }.intersection(self.moves)
-                
+
                 """
                 Check if the pokemon has abilities that can grant moves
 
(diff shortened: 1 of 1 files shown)


hsahovic:
Thanks @SantiagoRR2004 !

PR #803, opened 2025-12-09
If hp status says no status, overwrite status to None (plus a couple stray bugfixes)

cameronangliss:
One case this can matter is when a pokemon is revived, in which case their hp will be reset to a >0 value but their status would not have been reset from FNT to None.

### DIFF
--- a/src/poke_env/battle/abstract_battle.py
+++ b/src/poke_env/battle/abstract_battle.py
@@ -1009,9 +1009,9 @@ class AbstractBattle(ABC):
             mon.terastallize(type_)
 
             if mon.is_terastallized:
-                if pokemon in self.team.values():
+                if mon in self.team.values():
                     self._used_tera = True
-                elif pokemon in self.opponent_team.values():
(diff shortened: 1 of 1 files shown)


cameronangliss:
@caymansimpson I had to change a test case you had in your gen 9 calc after making this change. Can you check if your test was not correct? If it is correct, I'll have to go deeper into the code to figure out why this is happening, but I just want to verify with you that this test is actually solid. My code change is a one-liner, feel free to ask any questions if you need better clarification on what I'm doing here.

cameronangliss:
Update: I spoke with @caymansimpson and he's alright with the changes. This is ready for merge on my end👍

hsahovic:
@cameronangliss can you rebase / fix the conflicts? Otherwise LGTM

cameronangliss:
@hsahovic done, and the unit tests made me realize I had a couple of small bugs from the PRs you just pulled in (my bad), so I fixed those too. Ready for you.

PR #805, opened 2025-12-09
Track regenerator healing correctly

cameronangliss:
To do this, we also must properly track the mummy and trace abilities (which are abilities that can change a pokemon's ability), as well as neutralizing gas, which is better tracked as a field than an effect, since when active it affects all active pokemon, even if a pokemon switches in after it has activated.


hsahovic:
Thanks!

PR #812, opened 2025-12-13
Add ability tracking to strict consistency checking test

cameronangliss:
Just what it sounds like - we are now also tracking ability (both base and current ability) in the strict consistency checking. This PR includes all changes necessary for the tests to pass; I've ran them a several times myself and things look to be holding up. The most substantial tracking change is to the get/set methods of `ability`, which will return `_ability` unless `_forme_change_ability` exists, and then it will return that, UNLESS `_temporary_ability` exists, and then that will be returned above all else. Also, it is now impossible to overwrite a non-None ability with None as this never happens in Pokemon to my knowledge, but please correct me if I'm wrong.

### DIFF
--- a/src/poke_env/battle/abstract_battle.py
+++ b/src/poke_env/battle/abstract_battle.py
@@ -690,7 +690,7 @@ class AbstractBattle(ABC):
                     # |-ability|p1a: Gardevoir|Intimidate|[from] ability: Trace|[of] p2a: Luxray
                     if mon.temporary_ability is not None:
                         mon.temporary_ability = None
-                    elif mon._ability is not None:
+                    elif mon.ability is not None:
                         mon._ability = None
                     mon.ability = "trace"
                 mon.ability = cause


src/poke_env/battle/abstract_battle.py
```diff
@@ -672,24 +672,32 @@ def parse_message(self, split_message: List[str]):
...
+                len(event) > 5 and event[5].startswith("[from] ability: Trace")
             ):
-                self.get_pokemon(pokemon).set_temporary_ability(cause)
+                if mon.ability != "trace":
+                    # correcting for bad PS ordering of logs, eg:
+                    # |-ability|p1a: Gardevoir|Intimidate|boost
+                    # |-ability|p1a: Gardevoir|Intimidate|[from] ability: Trace|[of] p2a: Luxray
+                    if mon.temporary_ability is not None:
+                        mon.temporary_ability = None
+                    elif mon._ability is not None:
```
hsahovic: I think we should handle this without accessing a private attribute

hsahovic:
Thanks!

PR #821, opened 2026-01-14
Make `Battle`/`DoubleBattle` `pickle`-able

cameronangliss:
It's nice to be able to serialize/deserialize objects so you can store them in files and save your current situation. Now you can do that with Battle and DoubleBattle objects! Battle-related objects no longer store a _data field containing GenData, and instead just call GenData.from_gen whenever they need the pokedex or something. This does not lose any performance and we were already doing this in some places, but allows these objects to be pickled now that they aren't carrying a pointer to the giant GenData object.

### DIFF
--- a/src/poke_env/battle/pokemon.py
+++ b/src/poke_env/battle/pokemon.py
@@ -22,7 +22,6 @@ class Pokemon:
         "_base_stats",
         "_boosts",
         "_current_hp",
-        "_data",
         "_effects",
         "_first_turn",
         "_gen",


cameronangliss:
That was not accidental, that branch was just trying to do a little too much I thought, so I made a new branch that had only changes that felt more agreeable.

Regarding my confidence about performance, I only meant that I can tell by reading the code. The from_gen method grabs an already-made GenData instance and uses that. It is written this way because the data loading in GenData would be needlessly repetitive and costly otherwise. Also, we were already using from_gen in other places with no issues. Thus, there is no reason to believe that performance would suffer from this way of doing it.

EDIT:
I was looking at the DataGen class again to make sure I'm right about this and realized we were double cache'ing the DataGen objects by accident, so I fixed that: b9464e8fe07a96b4250f4684431721ee0df1334a

hsahovic (approved):

PR #831, opened 2026-03-05
Introduce MoveSet dataclass for move tracking infrastructure

cameronangliss:
Replace Dict[str, Move] with MoveSet container that supports Mimic and Transform move overrides. Add base_moves, mimic_move, and transformed properties to Pokemon. Remove >4 moves pruning logic.

Rewrite transform() to use temporary base stats and types instead of overwriting species data. Detect Imposter ability in -transform events. Clear temporary state on faint/switch_out. Fix gen<3 single-ability assignment to avoid triggering ability setter side effects.

Handle Mimic effect in -start and -activate events to track moves copied by Mimic. Sets the MoveSet mimic_move property when Mimic is detected.

### DIFF
--- a/src/poke_env/battle/move.py
+++ b/src/poke_env/battle/move.py
@@ -952,6 +952,7 @@ class MoveSet:
     def __setitem__(self, key: str, value: Move):
         assert not value._from_mimic, "Set mimic-copied move with mimic_move setter"
         self.base_moves[key] = value
+        assert len(self.base_moves) <= 4, "A pokemon cannot have more than 4 moves"
 
     def _resolved(self) -> MoveSet:
         """


src/poke_env/battle/move.py
```diff
@@ -915,3 +928,69 @@ def weather(self) -> Optional[Weather]:
...
+    _base_moves: dict[str, Move]
+    _mimic_move: Move | None = None
+    _transform_moves: MoveSet | None = None
+
+    def __getitem__(self, key: str) -> Move:
+        return self.moves[key]
+
+    def __setitem__(self, key: str, value: Move):
+        assert not value._from_mimic, "Set mimic-copied move with mimic_move setter"
+        self.base_moves[key] = value
```
hsahovic: should this check for len <= 4?

hsahovic (approved):

PR #891, opened 2026-04-10
Learnset and battle format of each Pokémon

SantiagoRR2004:
I am adding the property of the normal learnset of each Pokémon, so later we can more easily process _Zoroarks_. It is nice to store the format because some don't use the normal one.

### DIFF
--- a/src/poke_env/battle/pokemon.py
+++ b/src/poke_env/battle/pokemon.py
@@ -103,8 +103,9 @@ class Pokemon:
         self._ivs: list[int] | None = None
         self._nature: str | None = None
         self._format: Optional[str] = format
-        self._learnset_format: bool = not any(
-            f in (self._format or "") for f in self._ALL_MOVES_FORMATS
+        self._learnset_format: bool = (
+            not any(f in (self._format or "") for f in self._ALL_MOVES_FORMATS)
+            and gen >= 3
         )
 
         # Battle related attributes
@@ -676,7 +677,29 @@ class Pokemon:
         self._weightkg = dex_entry["weightkg"]
 
         # Now the moveset
-        learnsets = GenData.from_gen(self.gen).learnset
+        current_gen = self.gen
+        while current_gen >= 3 and not self._learnset:
+            self.update_learnset(species, current_gen)
+            current_gen -= 1
+
+    def update_learnset(self, species: str, gen: int) -> None:
+        """
+        Update the learnset of the Pokemon based on its species and the gen.
+
+        This function is used to obtain a non empty learnset for the Pokémon
+        by going backwards if it has been Dexited. The learnset is not
+        available for the Gens 1 and 2 so this function would do nothing
+        for those Gens.
+
+        Args:
+            - species (str): the species to update the learnset for
+            - gen (int): the gen to update the learnset for
+
+        Returns:
+            - None: this method updates the learnset in place and does not return anything
+        """
+        dex_entry = GenData.from_gen(gen).pokedex[species]
+        learnsets = GenData.from_gen(gen).learnset
 
         all_moves: Dict[str, List[str]] = defaultdict(list)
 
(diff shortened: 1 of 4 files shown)


SantiagoRR2004:
@cameronangliss, I was looking at the future of using the learnset and there would be conflicts on some tests. Some of the battles should be changed to anythinggoes or have so their moves altered for that generation. Is there any documentation about how they handled Dexit? Should the learnset be the one of gen 7?

cameronangliss:
hi @SantiagoRR2004, I'm not able to come up with a good answer after a quick lookover, I'll have to find some time to look at this later, maybe next weekend.

SantiagoRR2004:
OK. I did some research and in the conclusion of the following [Smogon article](https://www.smogon.com/articles/dexit-most-missed) it says that in some formats you can use the Pokémon that are not in the Pokédex. They simply have their moves. So I made it go back to previous generations and get the possible moves from there. I think this could be ready to merge.

src/poke_env/battle/pokemon.py
```diff
@@ -662,6 +675,77 @@ def _update_from_pokedex(self, species: str, store_species: bool = True):
         self._heightm = dex_entry["heightm"]
         self._weightkg = dex_entry["weightkg"]
 
+        # Now the moveset
+        current_gen = self.gen
+        while current_gen >= 3 and not self._learnset:
```
hsahovic: I think this should be in the function

src/poke_env/battle/pokemon.py
```diff
@@ -662,6 +675,77 @@ def _update_from_pokedex(self, species: str, store_species: bool = True):
...
+            self.update_learnset(species, current_gen)
+            current_gen -= 1
+
+    def update_learnset(self, species: str, gen: int) -> None:
```
hsahovic: I think this should be cached and handled by `GenData`

hsahovic (changes requested):

SantiagoRR2004:
Now the learnset is cached in `GenData`.

src/poke_env/battle/pokemon.py
```diff
@@ -93,6 +101,11 @@ def __init__(
         self._evs: list[int] | None = None
         self._ivs: list[int] | None = None
         self._nature: str | None = None
+        self._format: Optional[str] = format
+        self._learnset_format: bool = (
```
hsahovic, line 103: is this used anywhere?

src/poke_env/battle/pokemon.py
```diff
@@ -61,6 +64,9 @@ class Pokemon:
         "_weightkg",
     )
 
+    # The formats that have a different learnset
+    _ALL_MOVES_FORMATS = ["hackmons", "anythinggoes", "stabmons", "sketchmons"]
```
hsahovic: Why anything goes?

hsahovic (approved):
I left two minor questions - lmk what you think

src/poke_env/battle/pokemon.py
SantiagoRR2004: I have searched and I haven't found the reason why. Maybe I thought that it could be used in the future. It can be removed without the code breaking.

src/poke_env/battle/pokemon.py
SantiagoRR2004: I probably got confused. I will fix it in a commit.

SantiagoRR2004:
This last commit was an empty one because one of the checks failed. It was a TimeoutError and probably was GitHub's fault.

src/poke_env/battle/pokemon.py
hsahovic: Yep let's remove it!

hsahovic: After that I think we're good to go :)

SantiagoRR2004:
I think that is everything.

One curious thing about the [learnsets](https://raw.githubusercontent.com/hsahovic/poke-env/refs/heads/master/src/poke_env/data/static/learnset.json) is that they all start in generation 3. I made the learnset checker (`self._learnset_format`) only be True when it is higher than 2 because of this.

src/poke_env/battle/pokemon.py
SantiagoRR2004: I committed a fix that removed it.

hsahovic:
Thank you @SantiagoRR2004 !

PR #898, opened 2026-04-15
Support best-of-3 battles

cameronangliss:


src/poke_env/player/player.py
```diff
@@ -256,12 +266,54 @@ async def _get_battle(self, battle_tag: str) -> AbstractBattle:
...
+                    split_message[2] == self.username
+                    if split_message[1] == "win"
+                    else None
+                )
+                await self._battle_count_queue.get()
+                self._battle_count_queue.task_done()
+                async with self._battle_end_condition:
+                    self._battle_end_condition.notify_all()
+                self.ps_client._battle_locks.pop(game_tag, None)
+                if hasattr(self.ps_client, "websocket"):
```
hsahovic: why is this necessary?

hsahovic (approved):

src/poke_env/player/player.py
cameronangliss: The ps_client will only have a websocket if the player has start_listening=True

cameronangliss: I simplified things: 5e43091c45c512755ec1efe7603ee231c6e184a1

PR #921, opened 2026-05-31
Improve battle parsing performance

hsahovic:


cameronangliss:
Which parts of this do we get speedup? Some of this looks to me like refactor. In MoveSet for example, the changes don't immediately appear like they would give a performance improvement over the original. Or, I should say, the performance would be extremely minimal/rare. Oftentimes in scenarios like that I think leaving the code simple is better.

In other cases, it looks like we're foregoing using properties in favor of just calling the internal code of the property at the callsite. I'm open to being wrong but I'm not sure that's necessary.

hsahovic:
This is very much experimental at this point. Here is the summary codex gave me of the current state:

This PR improves Python-side battle parsing performance while preserving existing behavior. I benchmarked with a parse-only replay workload built from 20 captured gen9randombattle battles, replayed 100 times per trial for 2000 total battle replays, with battle requests and protocol messages fed back through the normal Battle/Pokemon parsing pipeline. The benchmark excludes simulator/network time and policy choice time, so it isolates Python state tracking and parser overhead.

Commit-level results on that workload were:
```
master:                                      5690 turns/s
0d611191 Optimize Pokemon request hot paths: 7118 turns/s  (+25.1%)
73a1bad7 Avoid repeated battle lookups:      7257 turns/s  (+2.0%)
23b6a9aa Reduce protocol parsing overhead:   7621 turns/s  (+5.0%)
```
The largest gain comes from the Pokémon request/move hot paths: avoiding repeated self.moves / MoveSet.moves resolution, bypassing _add_move for the common non-Transform/non-Mimic request path, faster HP parsing for common current/max strings, and reducing repeated property calls while building available moves. The battle lookup commit removes repeated active_pokemon, team, trapped, and fainted property calls in request parsing and switch handling. The protocol parsing commit avoids copying every protocol message, caches event type and team lookups in hot paths, and keeps copying only for the move branch that mutates the event. The final commit only inlines one-off parser constants for readability and was benchmark-neutral.

Let me know what you think @cameronangliss !

cameronangliss:
Okay, I guess that pretty much convinces me. I do think it's worth adding complexity to poke-env for the sake of user performance. Maybe we weren't using properties very responsibly to begin with. I still feel like it's a bit of sacrificing "best practices" persay in terms of readability, but I think we aren't losing much on that regard anyway.

hsahovic:
I agree that it's worth compromising for performance, to a point. The current state of this branch is close to my decision threshold.
I'll come back to it with fresh eyes and try to clean things up a bit - we might be able to keep most of the performance gains with a better diff.

PR #962, opened 2026-09-18
Fix Pursuit PP tracking against Pressure

hsahovic:
Adjust Pursuit parsing and PP tracking to match current Showdown output; the Gen 4 PP exception below matches an upstream bug and should be removed before merge.

Also recognize the newer `[from] move: Pursuit` protocol suffix as an interception rather than a move calling another move. Regression tests cover switching and non-switching targets in generations 3–5.

The Gen 4 exception currently charges only one PP when Pursuit intercepts a switching Pressure target. Cameron is correct that the game should charge two: [Platinum's Pursuit implementation](https://github.com/pret/pokeplatinum/blob/d18093e4fd79ed5260d98363a00f315441b6c656/src/battle/battle_script.c#L6915-L6923) explicitly deducts the extra PP. The existing poke-env behavior is correct for this case. Keep the protocol-suffix fix, remove the Gen 4 exception and its one-PP expectation, and fix Showdown upstream.

This PR changes battle tracking only. CI maintenance landed in #960; error propagation is in #961. Rebased onto master after #960 merged.

## Validation

- `uv run --no-sync pytest -q unit_tests/` — 340 passed on this branch.
- Black, isort, flake8, and mypy passed.


cameronangliss:
I am 99% sure this is a bug in showdown, and we actually are tracking correctly.

